PR #32, opened 2015-01-05
fix(Icons): OZF-734 fixed path issues for dist

wski:


rpokorny:
I figured out how we are handling the paths for the other fonts, and implemented that for the icon font instead of your fix.

PR #36, opened 2015-01-23
Global Toolbar

sdeighan:
I believe that the global toolbar can now be merged


nishp1:
Thanks, I will take a look.

Issue #133, opened 2015-03-17
Fix unit test webpack error 


wski:
Fix unit test webpack error;

wski:
Issue is resolved but need to test on other machine @ branch unittest-webpack-patch

PR #135, opened 2015-03-17
chore(Unit Tests): Fix webpack error in unit tests

wski:


wski:
apparently a issue is created if you create a commit with the word fix in it and no issue #

wski:
same thing as #133

Issue #167, opened 2015-04-03
Listing Management - The scroll bar moves from the right side of the browser to the right side of the content after editing a listing


mparizernc:
Develop Branch

I noticed that when I went to the listing management pages that there were 4 columns of listings. I then went to edit a listing and went back to the Listing Management pages and there was only 3 columns. After further investigation I found that the scroll bar initially is at the right side of the browser. After coming back from editing a listing the scroll bar moved from the right side of the browser to the right side of the content causing less usable space causing the reducing of the columns.  (Content means where the white background and gray background meet.)

rpokorny:
I think I know what's causing this.  The create/edit page is styled such that the page overall never scrolls - only the form in the middle does.  The only way that I could cause this behavior was to add (when create/edit is opened) a css class to one of the very high-level elements (like `html` or `body`) that changes their sizing.  This class is probably not being removed appropriately when the create/edit page is closed.

hoganmaps:
I can't seem to replicate the 4 to 3 column issue with the latest chrome or FF24, but i see the scroll bar part

mparizernc:
Interesting I was able to reproduce it on every browser. But I assume fixing the scroll bar should fix the column issue I was seeing.

hoganmaps:
Just made a change that should fix the scrollbars on develop

mparizernc:
All Browsers

Verified that the scroll bar no longer moves.

jloveland18:
All Browsers.
Verified that when I went to the listing management pages there were 4 columns of listings. I then went to edit a listing and went back to the Listing Management pages and there were still four columns.

PR #181, opened 2015-04-08
fix(listingManagement): When mounting create-edit, `create-edit-open` cl...

hoganmaps:
...ass was appended to main, but not removed, causing the side scrollbar to shift from body to the main class. #167

### DIFF
--- a/app/js/components/createEdit/index.jsx
+++ b/app/js/components/createEdit/index.jsx
@@ -474,10 +474,6 @@ var CreateEditPage = React.createClass({
 
     onClose: function () {
         this.transitionTo('my-listings');
-        var main = $('#main');
-        if(main.hasClass('create-edit-open')){
-          main.removeClass('create-edit-open');
-        }
     },
 
     onPreview: function () {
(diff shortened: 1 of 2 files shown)


app/js/components/createEdit/index.jsx
```diff
@@ -474,6 +474,10 @@ var CreateEditPage = React.createClass({
 
     onClose: function () {
         this.transitionTo('my-listings');
+        var main = $('#main');
+        if(main.hasClass('create-edit-open')){
```
rpokorny: FYI this check is redundant - removeClass won't do anything if the class isn't present

app/js/components/createEdit/index.jsx
```diff
@@ -516,13 +520,11 @@ var CreateEditPage = React.createClass({
...
     componentWillMount: function () {
         var main = $('#main');
-
         main.addClass('create-edit-open');
     },
 
     componentDidUnmount: function () {
         var main = $('#main');
-
         main.removeClass('create-edit-open');
```
rpokorny: This line should've prevented this problem from ever occurring.  It might be worth looking into why it didn't work.

PR #184, opened 2015-04-08
Issue 167

hoganmaps:
Dug a little deeper per comments from Ross on the last pull request. Close that and re-submitted a PR with the scrollbar fixed on the unmount part of the component lifecycle.


rpokorny:
So `componentDidUnmount` isn't a thing apparently... huh.

wski:
I guess that makes sense, if the component did unmount that means that the code you are trying to execute is undefined since it's been unmounted. whereas willunmount is executed right before that happens.

PR #241, opened 2015-05-11
fix(Changelog): Added fallback for legacy actions

wski:


wski:
merge to develop not master

Issue #258, opened 2015-05-15
Create/Edit Page - Add unavoidable message on Save errors


clarkjefcoat:
Two users are reporting saving edits on the Create/Edit page only to return to the website later to find that their edits were not saved.  When I try to reproduce these errors I get error messages on the save.

So either these users are not getting errors (and are unable to reproduce this situation) or the errors messages are being displayed but are escaping the attention of the users.

Can we add a popup or some other mechanism that is impossible to miss when attempts to save do not work due to errors?

sdeighan:
Here is a modal that we can display in the event that a save is not successful.

![mp_createedit_saveunsuccessful](https://cloud.githubusercontent.com/assets/7987446/7658386/bfe4b382-fb07-11e4-9b1e-a0e6972f180a.png)

clarkjefcoat:
Confirmed working on ci-master using FF24, Chrome, IE9.  (IE9 does not do animation, but does show message).

kgordon-1:
Tested on STG
Fixed, it now displays a big Red X and a Red bar over the error field.  The Red bar contains the error message.

PR #262, opened 2015-05-18
fix(style): Updating SweetAlert styling

sdeighan:
Added overrides for SweetAlert styling


sdeighan:
Update styling for #258

Issue #260, opened 2015-05-18
Detailed View - URL updates with every change (Legacy adaptor listings)


sdeighan:
1. Log into Marketplace as Admin
2. Pick a listing and open detailed view
3. Open in Create / Edit or admin tab
4. Change anything about ~~the listing~~ a legacy adaptor listing (?)
5. Open Detailed View, Admin Tab
6. View details of the change that was made
7. See interesting URL information

![detailedview_changesbug](https://cloud.githubusercontent.com/assets/7987446/7684627/84362dc0-fd55-11e4-9d7e-630e34d9b49d.png)

Changes made for this image were (in chronological order)
1. Flagging as featured
2. Changing version number to 1.23

(Also, after I did the second change, I opened the Featured flag change again and instead of being below the launch URL change like the version number change details, it's now above the URL change - as seen in the screenshot.)

mparizernc:
Looking at the screenshot this is accurate in the fact that the launch url is changing.  It appears that each time the listing is saved the % value gets encoded again for example if you have a % in the url it encodes with %25 and when you edit and save the listing it saves as %2525.

wski:
I had fixed this in #222, not sure how it was undone. I'll have a look at it though; I was able to replicate it.

clarkjefcoat:
Confirmed working on ci-master with Chrome, FF24 and IE9.

kgordon-1:
Passed – Tested in STG
Verified fixed by modifying fields in listings and adding spaces to the URL.  In the preview, the URL displays %20 and the changes to fields are displayed correctly.

PR #263, opened 2015-05-18
chore(Create Edit): only replace spaces with encoded spaces if spaces exist #260

wski:


rpokorny:
I said it when this issue first came up, but I'll say it again.  I strongly believe that the correct way to address this is to not allow spaces in URL fields at all.  No real-world URL will have spaces, since spaces are not allowed in URLs.  Doing validation of that would be much simpler than all this escaping/unescaping gymnastics.

wski:
Agreed, Clark and I talked for a while yesterday and we've came to the agreement that we would only correct their mistake if they added spaces and from them on only display the encoded url in hopes they realize the mistake. 

We've simplified it quite a bit because the hole we were digging by trying to decode and encode all over the place was very deep.

rpokorny:
That... isn't quite what I was saying.  I'm saying it should fail to validate if the URI is invalid.  Spaces are not valid characters in a URI.  Thus if they put a space in the URI text field it turns red and they can't save.  If they are trying to maunally type in a URI that should include %20s, they can (and should) encode it themselves

wski:
I understood that but from what I've been told it's a problem with our end user that happens a lot so we are trying to ease up on some of their confusion. I try to assume my mom is the same as an end user and I cannot see her figuring out how to encode a URL I think she would give up on creating all together

rpokorny:
I don't see low-tech end users writing URIs by hand (as opposed to copy/pasting them) but ok

Issue #301, opened 2015-06-26
Create/Edit Page - Error Pop-up shown when there are no errors


clarkjefcoat:
1. From the dropdown menu choose "Submit a Listing"
2. Fill-in a subset of the required fields
3. Press Save
4. Press Submit, you will get an error as you have not filled in all required fields.  This is correct.
5. Update one of the fields you have already entered.
6. Press Save.
7. You will get an error pop-up.  If you press Show Errors, you will not be shown any errors because there aren't any.  The last error pop-up should not appear.

mparizernc:
I have verified that the error message no longer appears when you save after trying to submit without all the necessary fields populated.  This passed for develop.

mparizernc:
Verified on Staging that I can save even after a submit fails with validation errors.  Works on all browsers.

PR #306, opened 2015-07-02
fix(createEdit): Disable error modal on saving drafts #301

blaino:
Error modal still in-effect for saving changes already submitted, but
since users can save work-in-progress that isn't yet submitted the error
modal shouldn't always pop up on save.


blaino:
@wski Please have a look when you get a chance.

Issue #284, opened 2015-06-11
Create/Edit Page unable to delete the last character in What's New field


kgordon-1:
The last character in the What's New field will not allow you to delete it.

blaino:
The first comment here refers to a first unsuccessful attempt to fix this problem. That solution failed on computers with less cpu/memory. Adding lots of text to a textarea field and holding down the delete caused unpredictable jumps, skips and reintroduction of deleted data. All of that said, the idea behind this unsuccessful solution I believe is still sound: a form with validation shouldn't warrant React optimization hints with shouldComponentUpdate().

wski:
verified it's working on my end, passing to CI-QA

mparizernc:
I was able to delete all of the What's New with no chars left behind.  I tried before saving the value and after saving the value.  Works on staging in each browser.

PR #310, opened 2015-07-07
fix(createEdit): Prevent undeleteable characters in optional fields #…

blaino:
…256 #284

This fix does not address the root cause of the problem which is still
unknown. By doing the validation and not the update in
CurrentLIstingStore.js's onUpdateListing() the problem with the
undeleteable character in optional fields goes away.


blaino:
@wski Please have a look.

PR #315, opened 2015-07-14
fix(createEdit): Prevent intermittently blank Listing Type

blaino:
Update the select2input fields like Listing Type after a delay to allow time for API to return with options. For some unknown reason, ListingForm's shouldComponentUpdate() is not picking up on changes to this.props.system.types despite its use of lodash's isEqual() which is supposed to do a deep comparison. 

Note that testing this is difficult. I have found that you can increase your chances of recreating the problem if you rebuild (npm start), wait around 2 minutes, and refresh.


blaino:
@wski Please have a look.

wski:
Will revisit this in the future for a better fix. Merging for now

Issue #266, opened 2015-05-20
Create/Edit Page: In Chrome: Listing Type drop down menu is blank


kgordon-1:
Creating a listing, the 1st time, the Listing Type drop down menu is blank.  

It does not happen all the time but I notes it when I first open Chrome and go to the Create a Listing page.

sdeighan:
When this happens, is the initial selection blank? Or is the whole list blank?

kgordon-1:
The whole list is blank.

clarkjefcoat:
If I start in HUD or webtop and then choose "Submit a Listing" from the dropdown, the Create/Edit page ListingType dropdown is blank.  If I start in Marketplace and choose "Submit a Listing" the resulting Create/Edit page looks fine.

wski:
Have we been able to replicate this?

clarkjefcoat:
I still see this on ci-master in FF 24.0 with Ubuntu 14.04 when I click on "Submit a Listing" from HUD:

![image](https://cloud.githubusercontent.com/assets/7015118/8341287/e3482f26-1a91-11e5-8a74-d404da621665.png)

blaino:
@wski, Please test this when you get a chance - the perfect opportunity for you to try BrowserStack.

mparizernc:
@blaino I am not sure if this is because of this change but I am having routing issues now when trying to navigate away from the Create Listing page that I was not seeing before. 

Steps;
1 Navigate to the Create Listing page with Submit a listing
2 Select the listing management link next to the buttons
3 You remain on the Create page
4 Refresh and you are taken to the Listing Management page

Error in console:
[...]

PR #316, opened 2015-07-15
fix(createEdit): Prevent routing issues caused by original fix for #266

blaino:
@wski Please test locally. I believe I have solved the issue found by @mparizernc and the issue you found where getDOMNode() is undefined.


wski:
I'm worried about how this will work in prod. with the reported 20s+ response times

Issue #293, opened 2015-06-18
Create/Edit Page - For short description show a countdown for the number of characters remaining.


clarkjefcoat:
When entering the short description show the number of characters remaining so submitters know how long their descriptions can be.

mparizernc:
@wski This should stop the user typing at 0 not go into the negative.

wski:
We came to the conclusion to leave the negative so that the user can see how many characters they are over by. The same functionality exists in Twitter, Facebook and many other platforms. However if @bethany and @sdeighan requested that it stop at zero its pretty easy to switch over

sdeighan:
It's considered a best practice with form fields to only allow valid input wherever possible. I would strongly support just putting a limit on the text input field at the maximum number of characters. @bhalt Would you agree?

wski:
If they were to copy paste from somewhere though it might be harder for them to figure out how much they need to trim off if it just says `0` was my only thought

bhalt:
I'm good with leaving the negative as long as there is an error message after they go over 100, which there is. I agree with Matt about copying and pasting from elsewhere and using the text field to crop it down. I've done this before and it's very nice to be able to do it in the native environment with a live countdown, rather than having to open another program.

sdeighan:
If we want to do that, it's fine, but then I would suggest we redo the message. Negatives out of 100 is pretty confusing. I would either continue the count and have it stay red or switch the text to (# characters over maximum of 100.)

It's not possible to have a negative number if there's text in the box and, while I understand what you were going for, it may lead users to believe that there's an error with the counting mechanism.

bhalt:
I just now saw the screenshots (I'm not at a place I can build). Agree with Sarah -- let's change the message under the form field to "__ characters remaining" when under the limit and a red "__ characters over the limit" when over the limit. Thoughts @sdeighan?

Also, can the red highlight above the field just be on the "Must be 100 characters long or less" text when a user goes over the limit, and not also on the "May contain letters, numbers, or symbols" ?

sdeighan:
We had the same concern, either one sounds good to me.

The modified red highlight makes sense to me - I'd be all for it. I'm just not sure about the technical side, those errors are tricky unless they've been re-written recently. If we can get that changed, I'm totally on board.

jal-vsi:
Verified pass on all 4 browsers using testUser1: Error message if you type above 100, won't save >100, saves at 100 or less

mparizernc:
[...]

PR #317, opened 2015-07-15
feat(CreateEdit): Added character limit for short description #293

wski:
@arw180 @blaino @clarkjefcoat can you review


app/js/components/createEdit/form/InputMixin.jsx
```diff
@@ -13,13 +13,16 @@ var InputMixin = {
 
     render: function () {
         var labelClasses = classSet({ 'input-optional': this.props.optional });
-
+        var CharLimit = (this.props.charLimit && this.props.value) ? CharLimit = `${(this.props.charLimit - this.props.value.length > 0) ? this.props.charLimit - this.props.value.length : 0 }/${this.props.charLimit} characters remaining.` :  `${this.props.charLimit}/${this.props.charLimit} characters remaining.`;
```
blaino: too long.

app/js/components/createEdit/form/InputMixin.jsx
```diff
@@ -13,13 +13,20 @@ var InputMixin = {
 
     render: function () {
         var labelClasses = classSet({ 'input-optional': this.props.optional });
-
+        // if we have a charlimit and value is not undefined display the characters remaining.
+        var CharLimit = (this.props.charLimit && this.props.value) ?
```
blaino: This is fine, but my personal preference is to do the math first to get the number of remaining characters and just build the string after that.

app/js/components/createEdit/form/InputMixin.jsx
```diff
@@ -13,13 +13,20 @@ var InputMixin = {
...
+            `${this.props.charLimit - this.props.value.length}/${this.props.charLimit} characters remaining`:
+            // if we do have a character limit, but value is undefined set it to the char limit max remaining
+            `${this.props.charLimit}/${this.props.charLimit} characters remaining`;
```
blaino: Isn't this just "1 character remaining?" this.props.charLimit/this.props.charLimit = 1?

blaino:
:+1: 

I pulled it down and ran it locally - works as expected. 

@arw180, next review please?

arw180:
Please `rebase -i` these 5 commits down into 1. Also, use `feat` instead of `chore` for the commit type so that it shows up in the changelog. Further review pending ...

arw180:
This did not work quite as expected. When I type one char, it says 99 characters remaining (good so far), but when I type the second character, it still says 99 characters remaining. Typing a third character results in "98 characters remaining", and then it counts down from there.

wski:
for whatever reason using state instead of props seems to have rectified the issue fore me (could only reproduce in Safari, so tested there)

arw180:
After the last change it appears to work as expected, but your rebase wasn't done properly as there are still two commits

wski:
I am aware, I will reply here when I am ready for you to check

wski:
rebased to one commit

arw180:
:+1:

Issue #274, opened 2015-06-04
URL is in the Preview > Details view is truncated, leave full url and wrap to new line when too long


hoganmaps:

hoganmaps:
![uploads/a0023130-38ee-4afe-bf21-2de1ffcd4394/Screen Shot 2015-06-04 at 5.36.14 PM.png](https://s3-us-west-2.amazonaws.com/prod.huboard.com/uploads%2Fa0023130-38ee-4afe-bf21-2de1ffcd4394%2FScreen+Shot+2015-06-04+at+5.36.14+PM.png)

mparizernc:
@wski This does not appear to be working in IE. Looked fine in FF and Chrome.

![image](https://cloud.githubusercontent.com/assets/7975509/8775885/8c1a16b4-2eba-11e5-8cbe-953984c5b87d.png)

wski:
PR https://github.com/ozone-development/ozp-center/pull/336 should fix the IE issue

jal-vsi:
Verified working (testUser1) in all browsers (IE9, IE11, FF24, CH36)

mparizernc:
Verified that the URL looks correct and is word wrapped on all of the browsers on Staging. (IE9, FF, Chrome)

PR #321, opened 2015-07-15
fix(Detail View): Truncation to wrapped link #274

wski:


wski:
@clarkjefcoat and @blaino please have a look

arw180:
User facing, please change commit message to use `fix` instead of `chore` so we get it into the changelog

blaino:
Pulled. Tested locally with chrome. Works as expected. :+1: after fixing the commit message.

wski:
closing because I fixed this, along with another instance. This will only cause merge conflicts

PR #323, opened 2015-07-16
Added reporting of agency information when application launches are r…

gziskind:
…eported

This change will allow the metrics application to be aware of the creating agency when an application is launched.


wski:
:+1:

clarkjefcoat:
Will we also be adding tickets to track when applications are launched from HUD or Webtop?

Is there a corresponding ticket that goes with this pr?

wski:
@gziskind can you create a issue for this so that we can track it

gziskind:
sure #330

clarkjefcoat:
:+1:

Issue #324, opened 2015-07-16
createEdit: preview shows "Not provided!" for What's New when actually provided


blaino:
The fix (db76913) for #256 and #284 introduced a regression where the preview for What's New is not updated.

jal-vsi:
Verified working (testUser1) in both Create/Preview and Listing Mgmt/Preview for all browsers (IE9, IE11, FF24, CH36)

mparizernc:
Verified that the What's New text appears and is removed on the Details tab of the Detail View on the Create/Edit listing page before and after saving as well as on the Detail View on Listing Management. Tested on Staging with IE9, FF, Chrome

PR #325, opened 2015-07-16
fix(createEdit): Maintain preview without undeleteable chars #324

blaino:
Commit db76913 fixed the undeleteable character, but introduced a bug
where preview of the listing is no longer updated for What's New. Use
lodash's isEqual to do a deep comparison of props; update component if
there is a change for anything in props. Do this instead of modifying
_onChange() which, as evidenced by this bug, is more dangerous.


blaino:
@wski Need the first review. Since this monkeys with shouldComponentUpdate() a bit, please test on you lesser machine to check for performance problems.

wski:
You spelled engineering marvel wrong in your last comment, but other than that I tested on my amazing machine on Chrome and Slowfari (Safari) and everything checks out :+1:

blaino:
@arw180, need second review, when you get a chance.

arw180:
didn't repeat tests that others did, but code and commit look good :+1:

PR #326, opened 2015-07-16
Fix empty listing type

blaino:
This change isolates the delay to select2input fields only - tag inputs aren't changed. @wski, please take a first look.


wski:
Is there a reason for merging master in verses rebasing master into your branch? If not please rebase to avoid multiple commits.

blaino:
@wski , No, no good reason. Should have rebased master into my branch. After talking with Alan, we've decided it's not worth trying to fix - as the fix would lose this conversation's history.

wski:
Verified that the tag fix works, unnecessary cancel timeouts and variables were removed. Everything looks :+1:

blaino:
@clarkjefcoat , need a second reviewer.

clarkjefcoat:
:+1:  (Could not test in IE when choosing "Submit a Listing" from the HUD dropdown due to an unrelated bug.  Otherwise, it looks good.)

clarkjefcoat:
This fixes #328 as well.

PR #331, opened 2015-07-21
feat(discovery): Preliminary integration of discovery page OZBE3

blaino:
Discovery page loads with images, provided center is built against
concomitant ozp-react-common change. Notifications are present but not yet
deleteable. Quickview comes up but with no real content as it requires
additional DRF endpoints. Note that this commit also includes an update
to the unit test reporting to make the output more human-readable.


blaino:
@arw180 Please have a look.

app/js/webapi/Listing.js
```diff
@@ -165,7 +167,7 @@ var ListingApi = {
     getOwnedListings: function (profile) {
         var id = profile ? profile.id : 'self';
 
-        return $.getJSON(API_URL + '/api/profile/' + id + '/listing')
+        return $.getJSON(API_URL + '/api/self/profile/' + id + '/listing')
```
arw180: will require a trailing slash here

arw180:
I like it. :+1:

Issue #334, opened 2015-07-21
Search - A shared search link with only orgs (or only listing type) does not work


clarkjefcoat:
Shared search links such as:
- .../center/#/home///Web%20Application/
- .../center/#/home////Test%20Organization

always report "No results found" even when there are results.

An ugly workaround to get results is to type and then delete something in the search text box.

jal-vsi:
Shared search (Search Results Share) passed in Master with all 4 browsers and 3 users: User, Org Steward, and Admin.

mparizernc:
Verified that the Search is passed correctly when searching for an Org or Type with a shared link.  Verified in IE10, FF and Chrome on Staging.

PR #335, opened 2015-07-21
fix(Discovery): Makes searches with only orgs return results

clarkjefcoat:
Issue #334


clarkjefcoat:
@arw180 - I'd like to get this in tomorrow's build.

arw180:
Tested locally - confirmed it fixes issue #334. Can @blaino or @wski take a quick look and do a visual inspection?
:+1:

blaino:
I pulled and tested. The no results found goes away, but "/home///Web%20Application/" yields a search for "Web Application" whereas "/home/Web%20Application/" yields a search for "Web" or "Application" - one gray label as opposed to two. Also, "/home///Web%20Application/" doesn't put anything in the search box so you can't clear it out with by clicking X. 

I'm wondering where these triple slash urls are coming from in the first place.  When I click share I get a link like: "/home/air%20bread///" not  "/home///air%20bread/". "/home/bread%20mail///" correctly yields a search for "bread" or "mail" on ci-develop currently. 

I must be misunderstanding something.

blaino:
Clark straightened me out. Three slashes: /<category>/<search>/<web application>/<organization>. I get it now.

wski:
I'd like to pull this down locally before deploying it, the removal of some initial data setting and moving stuff to componentWillUnmount seems weird. If we are relying on the componentWillUnmount to display data on that component that would be bad. Will talk with @clarkjefcoat and make sure I fully understand the changes

wski:
:+1: confusing git diff got me.

PR #336, opened 2015-07-22
fix(DetailView): Force wrapping for long strings and links in detail view

wski:


wski:
@clarkjefcoat @arw180 please have a look

PR #338, opened 2015-07-23
feat(bookmarks): Bookmark toggling works in discovery OZBE8

blaino:
OZBE8 = ozone-development/ozp-backend#8


blaino:
@arw180 have a look. 

So the ticket reads "Bookmark listings works on Discovery page". I have taken this to mean toggling the bookmarks on and off, and being able to confirm that change either via swagger or hud. Maybe there is some other aspect of bookmark "listing" that I don't yet know about?

rpokorny:
What change in behavior results from these changes?  It seems to me like this code would effectively do the exact same thing as it did before.

blaino:
@rpokorny this is a refactor to work with the Django Rest Framework backend. I'm hoping the behavior does not change. (This is not a pull request into master, but into the new-backend branch).

rpokorny:
OK, I see it now.  The important part is that you match on `x.listing.id` but then retrieve (effectively) just `x.id`, right?

arw180:
tested :+1:

blaino:
@rpokorny , right.

Issue #327, opened 2015-07-16
Indicate to metrics if a search included additional parameters


clarkjefcoat:
Based on the discussion in #291 we would like to indicate to metrics users when a search term was used in combination with additional categories, orgs, or types.

One proposal is to report "searchterm (+)" for searches that included other information instead of just "searchterm".

tonyalletag:
This has been deployed to production

PR #347, opened 2015-08-04
fix(SearchMetrics): Indicate to metrics where searches are combined w…

clarkjefcoat:
…ith other terms #327


wski:
:+1:

hoganmaps:
:+1: tested all three options in chrome and everything worked as expected

PR #358, opened 2015-08-13
Making center responsive to tablet size

bhalt:


bhalt:
@arw180 This is the new responsive Center changes. I didn't have a wide range of browsers / roles to test with on the VM I build/develop on, so would like to do a lot more testing once this is somewhere where it can be accessed.

arw180:
Great! Let's get a few folks to test this out on different browsers/different screen sizes over the next few days, and plan to merge this in *after* the next release (8/18), giving us a full week to test it out on ci-develop before these changes make their way to our staging environment

arw180:
:+1: Merging this in now so we get a solid week of testing before the next release

PR #359, opened 2015-08-19
chore(508): Change responsive colors to be 508 compliant (AA)

bhalt:


app/js/components/management/user/index.jsx
```diff
@@ -8,7 +8,7 @@ var _ = require('../../../utils/_');
 
 // component dependencies
 var NavBar = require('../../NavBar/index.jsx');
-var Header = require('../../header/index.jsx');
+// var Header = require('../../header/index.jsx');
```
wski: What was the reason for removing the header file, was it just changed to NavBar?

wski:
It looks like a few `.ext~` files or, squiglets as I've just named them got added in. We should try to add this to the `.gitignore` somehow.

wski:
It looks like Reminders were added to the sidebar in this commit as well, we should make a separate feat commit for this so that it gets added to the changelog.

app/js/components/management/user/index.jsx
bhalt: Not sure why it shows it as deleted and readded, but it's still there ... line 11

bhalt:
Yikes .. I may also be able to change my text editor to not save backups .ext~

wski:
@bhalt It might be a good idea for us just to add `*~` to all of our `.gitignore` files for our project too though. Others might have the same issue and that would prevent it.

app/js/components/management/user/index.jsx
wski: It looks like it was just commented out

bhalt: Oh, right, that's correct -- I removed the header from all Listing Management pages since it wasn't in Create/Add Listing anyways. Streamlining the header to only be used from the Discovery piece.

wski: Ok, cool

arw180:
I just rebased this PR against the latest `master` and removed the `~` temp files. Hopefully everything still looks as expected.

arw180:
:+1:

clarkjefcoat:
:+1:

PR #360, opened 2015-08-20
    feat(quickview): Updates for Quickview less Reviews tab OZBE21

blaino:
OZBE18 = ozone-development/ozp-backend#21

    As of this commit, trying to convert API responses to the existing
    format as soon as the calls to the API return - as opposed to using the
    format prescribed by the new API. The existing format in center is camel
    case, and humps (https://github.com/domchristie/humps) is used to
    convert keys in the listing object from python style lower case
    underscore to camel case.

    Note that this change is not just for Quickview - it affects
    everything that needs a listing: discovery, listings management, and
    search.


blaino:
@arw180 please have a look when you get a chance.

arw180:
humps. :+1:

PR #362, opened 2015-08-20
fix(Search Description): added and where nessecery between labels

wski:


wski:
@clarkjefcoat @hoganmaps please have a look when available

clarkjefcoat:
Multiple agencies should say "or".  Multiple listing types should say "or".

wski:
@clarkjefcoat new commit fixes this

clarkjefcoat:
Where you use "or" I don't think you need to make it plural.

image

Shouldn't the above say "with the type Web Application or Widget in the organization ... with the categories ...".  Change "types" to "type" and "organizations" to "organization".  But leave "categories" as is.

Also why isn't there a space in front of "in" and in front of "with"?

wski:
@clarkjefcoat

clarkjefcoat:
:+1:

hoganmaps:
:+1: this is acceptable

PR #365, opened 2015-08-20
fix(FireFox): Disabled Bootstrap animations if the user is using FireFox

wski:


wski:
@arw180 @clarkjefcoat I tested this on my firefox... But I don't really have a way to truly test that this resolves the issue because I cannot replicate it.

arw180:
I don't see any difference in FF between this change and master. In particular, I was hoping to see the 'slide' effect removed when you hover over an app tile on the Discovery page, but that's still there. Did you see certain animations removed with this change?

arw180:
Modal animations/transitions are another good target for removal, as those are also painfully slow in production on FF

wski:
I had only disabled jQuery animations from transition.js in BS. I just
pushed a change to disable all css3 transitions in firefox

On Fri, Aug 21, 2015 at 11:52 AM, Alan <notifications@github.com> wrote:

> I don't see any difference in FF between this change and master. In
> particular, I was hoping to see the 'slide' effect removed when you hover
> over an app tile on the Discovery page, but that's still there. Did you see
> certain animations removed with this change?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/ozone-development/ozp-center/pull/365#issuecomment-133469980>
> .
>



-- 
*Matt Wisniewski* \ *Software Engineer *

*a software engineer with a love for the web, JavaScript and bold coffee
overpowered by cheap creamer. *
http://wski.github.io
http://github.com/wski
<http://www.8bit.ws/>

arw180:
That looks better (well, it looks worse, which in this case is what we're shooting for). :+1:

clarkjefcoat:
:+1:

Issue #343, opened 2015-07-31
Submit a Listing bug


jal-vsi:
When submitting a listing the text in "Usage Requirement" does't wrap when viewed in Details (text does wrap in "What's New") as noted in the attached file.
![submit listing](https://cloud.githubusercontent.com/assets/13243617/9014300/9c2c0624-378f-11e5-9485-04a0c46445af.PNG)

jal-vsi:
Passed testing in Latest (Develop) using IE11, FF24, CH36 for new Listing submit.

tdressel:
Passed on Staging using Chrome 36, FF 31, IE 10

PR #368, opened 2015-08-24
fix(DetailView): Usage Requirements now wraps long text with no spaces #343

wski:


wski:
@clarkjefcoat @blaino

clarkjefcoat:
:+1:  This force wraps in the same way as What's New

blaino:
So we want to cut words right in the middle? 
screen shot 2015-08-25 at 10 04 44 am

wski:
This is the only reliable way to break super long strings that I could find. Though it is odd that it's wrapping words that could be brought to the next line, it's the only solution I could find that was compliant with all browsers

blaino:
I'm guessing that a more subtle approach could be found and we should look for it, but this gets rid of immediate ugliness. :+1:

PR #388, opened 2015-09-23
New backend categories

wski:


wski:
## Ready for testing @arw180 @blaino

PR #397, opened 2015-10-02
fix(image upload): Add dummy upload to make image upload work for IE11.

blaino:


rpokorny:
Did you try forcing IE11 through the IE9 code path (eg, doing the form-data via an actual form)?  Since <form>s have been around for much longer than the JavaScript FormData object, that path is likely to have fewer browser bugs.

blaino:
I did do a quick try of that but cloneNode() wasn't defined or supported or some such for the input to the IE9-handling function.

rpokorny:
cloneNode() has been supported in all browsers since forever... are you sure you passed in the \<input\> element to the ImageApi.save method, instead of the File object?

blaino:
No, not sure, but since we won't be supporting IE9 when this new backend hits production, I'm not sure keeping the hidden form and iframe solution around is a good idea.

rpokorny:
IMO its better than doubling the number of REST requests that are necessary

Issue #403, opened 2015-10-09
Listing Management Table View - Javascript error is displayed when missing owner data


clarkjefcoat:
Depending on the data loaded, you may see

Uncaught TypeError: Cannot read property 'tolowerCase' of null

in your console when loading the table view tab of some organizations in Listing Management.

clarkjefcoat:
This is caused by Owners who have a null displayname in the database.  We do have data like that in our production environment.

jal-vsi:
Passed testing in Master (IE9/11, FF, CH).  Admin creates draft listing (without owner org), Admin selects Menu/Listing Management, From Table View ensure Column Settings include Organization, Select Organization column and confirm draft listing is first (null Org listing).

PR #405, opened 2015-10-09
fix(ListingManagement): Table view does not work if the owner display…

clarkjefcoat:
…name is blank #403


clarkjefcoat:
@hoganmaps @blaino

clarkjefcoat:
@wski

wski:
:+1:

Issue #267, opened 2015-05-21
Create/Edit Page - Contact mislabeled after Save


clarkjefcoat:
On the Create/Edit page the labels on the Contacts are sometimes incorrect after pressing save.

1  Start by adding multiple Contact Types using Marketplace Settings

![contacttypes](https://cloud.githubusercontent.com/assets/7015118/7750035/e1297f0e-ff9f-11e4-864a-4abc22191899.png)

2  Create a new listing and add only the minimum amount of information

![2_create_new_listing](https://cloud.githubusercontent.com/assets/7015118/7750060/f95c4714-ff9f-11e4-82bf-7260f081e844.png)

3  Add multiple contacts.  For my test I named my contacts the same as the contact type (e.g.  the Support POC is named "Support POC", the Project Sponsor is named "Project Sponsor").  I added four contacts - only 1 is shown below.

![7_project_sponsor](https://cloud.githubusercontent.com/assets/7015118/7750097/387416f2-ffa0-11e4-936c-e18c9c164150.png)

4  Press Save.  Scroll down and review the contacts.

![8_bad](https://cloud.githubusercontent.com/assets/7015118/7750110/55d10264-ffa0-11e4-9af4-1087471a4a4a.png)

In this case the Support POC information is being shown under the "Project Sponsor" label.  This is the issue that needs to be fixed.

sdeighan:
@clarkjefcoat Did you mean to mark this as "working" or did you mean "ready"?

arw180:
tried this multiple times w/ 4 different contact types. did not see this problem occur.

kgordon-1:
Tested in STG
I have never had this situation to happen but did follow the steps above to re-create it.  Not reproducible.

clarkjefcoat:
This bug is back.

jal-vsi:
Passed testing in Listing environment using IE11, FF24, CH36.  Admin creates New Contacts, then creates a draft listing (1st two fields populated, adds multiple contacts, saves, checks for Contact Type/Name changes (none observed).  User creates draft listing and adds Contacts, no problems observed.

jal-vsi:
Passed testing in Release environment with same results as Listing.

tdressel:
Passed testing on staging using Chrome 36 , FF 31, IE 10.

PR #421, opened 2015-10-23
fix(create/edit): Add unique keys to each contact field #267

blaino:
@wski, @arw180 please have a look.


wski:
Looks :+1: to me, I think we should investigate https://www.npmjs.com/package/flakeid in the future for usage with keys in react as discussed.

arw180:
:+1:

Issue #423, opened 2015-10-26
Eliminate warnings


blaino:
Multiple attempts to upgrade Center's React version from v12 to v13 or v14 have failed. Facebook's upgrade documentation hints that the elimination of all warnings and errors will result in a smoother upgrade. This ticket is to tamp out the warnings. 

Note that most of the warnings in Center have to do with arrays in components not having unique keys. Another common warning is the deprecation of transferPropsTo(). 

Also note that there is a known error with create/edit preview (#422).

tdressel:
Tested on Latest - Chrome
Center | Help - Help Modal opens without error and has Help Modal content.
Center | Org Listings | Allowed user to Submit New Listing no error.

More testing pending.....

tdressel:
More Testing on Latest - Chrome 36, FF 24 and IE11
Help Modal, Org Listings, Detailed Query ( Types, orgs and cat), Listing Tiles, Sidebar, Recent Activity Sidebar, MyListings,Overview, Details, Reviews, Resources, Administration

tdressel:
Passed on Staging.

PR #424, opened 2015-10-26
fix(center): Eliminate (many) warnings #423

blaino:


app/js/components/discovery/index.jsx
```diff
@@ -239,7 +239,7 @@ var Discovery = React.createClass({
         }
 
         return (
-            <FeaturedListings
+            <FeaturedListings key="FeaturedListings"
```
rpokorny: `key` is only required when React elements are used within an actual JavaScript list.

app/js/components/management/shared/ApprovalStatusFilter.jsx
```diff
@@ -20,8 +20,8 @@ function filterOption (currentValue, label, value, count, htmlFor, className, ic
     }
 
     return [
-        <input id={htmlFor} type="radio" value={value} />,
-        <label htmlFor={htmlFor} className={className}>
+        <input id={htmlFor} key={"input" + htmlFor} type="radio" value={value} />,
```
rpokorny: `key` should not be required here

blaino:
It would be nice to get rid of these warnings but I can't, for the life of me, figure out how these changes broke the unit tests. Some of the failing unit tests, if you isolate them with describe.only will pass (e.g., the tests in EditReview-test.js). Sigh. I'm giving up. I guess we're stuck with React 12.2 forever. Either that or rewrite Center. (End: hyperbolic whining).

PR #426, opened 2015-10-26
feat(Table View Refactor): Refactored Listing Management section for …

KageRyu7:
…better organization, refactored table view to use w2ui, added export to excel option in table view, created a seperate unpaginated listing store for tableview, added a call in ListingsService.js to retrieve tableview's listings.


blaino:
In the future, please consider submitting separate pull requests for separate features/refactors. It's pretty hard to see which changes go with which feature/refactor. There's definitely enough going on here to warrant separate PRs.

app/js/components/management/shared/TableView.jsx
```diff
@@ -0,0 +1,331 @@
...
+        });
+
+        if (CSV === '') {
+            alert("Invalid data");
+            return;
+        }
+
+        var fileName = "ListingReport_" + ReportTitle.replace(/ /g,"_");
+
+        var uri = 'data:text/csv;charset=utf-8,' + escape(CSV);
```
blaino: jshint warning on this line. Also, check out: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/escape.

blaino:
The w2 table looks really good.

app/js/components/management/shared/TableView.jsx
```diff
@@ -0,0 +1,331 @@
...
+                toolbarEdit: false,
+                toolbarDelete: false,
+                toolbarSearch: true,
+                toolbarReload: false,
+                toolbarColumns: true,
+                toolbarSave: true
+            },
+            buttons: {
+                save : {
+                    caption: w2utils.lang('Export to Excel'),
```
blaino: This should say 'Export to csv'.

blaino:
My editor is showing some trailing whitespace in a few places. No big deal. If it's easy might as well get rid of them.

blaino:
When you select line number from the column editor, the line numbers don't line up nicely with the rows. I've showed this to Tony and Alan. We all agree it is something to fix, just not on this branch and not on this ticket, as we want to get this PR into today's build. Please write a new Center ticket to fix this, and if you have the time, fix on a separate feature/fix branch. 

screen shot 2015-10-27 at 10 54 17 am

blaino:
Ok, we have some known issues here, but we're going to initiate the sequence and merge. :+1:

arw180:
Works for me :+1:

blaino:
@jal-vsi found that export does not work in IE11. I tried with the console turned on: "The data area passed to a system call is too small".

bhalt:
can this branch be deleted @blaino @arw180 ?

blaino:
I think so.

Issue #428, opened 2015-11-03
Require Description when rating an App


tonyalletag:
A description should be required of > 20 characters when reviewing or rating an application in the center.

blaino:
Requirement from Tony: character limit needs to be in the config file.

tdressel:
Tested and passed on Chrome 36, FF 24 and IE 11.
Added and Edited Reviews with and without descriptions.  Warning if under 20 characters and allowed save if over 20 characters.

PR #433, opened 2015-11-09
feat(reviews): Require description when rating app #428

blaino:
@wski please have a look.


blaino:
Note that tests fail because of the parallel change in commons. There's a new config key there that center now depends on. All tests pass locally.

wski:
:+1:

blaino:
@clarkjefcoat, have a look?

clarkjefcoat:
:+1:

PR #441, opened 2015-11-23
(fix)Listing Type, Organizations dropdown and side Nav

SOAP1914:


SOAP1914:
@arw180 @wski @blaino

clarkjefcoat:
:+1:  I will take your word that this fixes JAWS issues.  It doesn't seem to have broken anything.

SOAP1914:
Thanks @clarkjefcoat lol Its alot of tabIndex I know. I've tested it on iOS VoiceOver and it worked fine.

wski:
I belive the tab index values are not correct long term, should use as follows

```
<SelectBox tabIndex="1">
   <Option tabIndex="2">Something Here</Option>
   <Option tabIndex="3">Something Here</Option>
   <Option tabIndex="4">Something Here</Option>
   <Option tabIndex="5">Something Here</Option>
</SelectBox>
```

If this is just for testing though, it should be fine. From what I can tell tab index defines  the order at which things are shown as you tab through an application with JAWS. With that in mind we should never have repeating tab indexes.

bhalt:
can this branch be deleted @SOAP1914?

SOAP1914:
yes

PR #442, opened 2015-11-24
(fix)Bookmark Button if its has a bookmark or not

SOAP1914:


SOAP1914:
@clarkjefcoat , @arw180, @wski, @blaino

clarkjefcoat:
:+1:  aria-label changes based on if the app is bookmarked

hoganmaps:
:+1: The aria-label is changed when you click the bookmark button on an app
screen shot 2015-11-24 at 6 13 45 pm
screen shot 2015-11-24 at 6 13 39 pm

PR #446, opened 2015-12-01
Add CasperJS groundwork for automated regression testing.

blaino:


clarkjefcoat:
:+1:

clarkjefcoat:
@blaino:  When capturing screenshots, can you do something to normalize the filenames created.  It can currently be ugly.  For example, when `waitAndClick("a[href='#/edit']", this);` captures a screenshot it creates a directory called "a[href='#" and saves an image called "edit'].png".

Can you remove the special characters and replace spaces with underscores (or something like that)?  That would make the filenames looks more reasonable.

blaino:
To make the filenames more reasonable I've added an optional parameter to the helpers that allows you to specify an alternate name for the file. (I've also made the png files save to a /captures directory so they don't pollute the top level).

hoganmaps:
this is cool and unobtrusive :+1:

PR #449, opened 2015-12-08
(fix)The modal nav on each app

SOAP1914:
To let JAWS nav thru the app modal (top nav)


SOAP1914:
@arw180 @blaino @clarkjefcoat @wski @hoganmaps

app/js/components/quickview/Header.jsx
```diff
@@ -30,12 +30,12 @@ var QuickviewHeader = React.createClass({
         return (
             <div className="quickview-header">
                 <div className="quickview-header-info">
-                    <button type="button" className="close" aria-hidden="true" onClick={ this.props.onCancel }><i className="icon-cross-16-grayDark"></i></button>
```
blaino: Why was aria-hidden ever "true"?

blaino:
Aside from my not understanding how aria-hidden is supposed to be used, LGTM. :+1:

app/js/components/quickview/Header.jsx
SOAP1914: lol np good question, i'll add to best practice (thanks)
In Bootstrap Modals...
The purpose of the aria-hidden on the close button is that in that markup the "X" or "times"-symbol won't tell screen readers much.

To make it accessible the close button should look something like this:

<button type="button" title="Close">
  <span aria-hidden="true">&times;</span>
  <span class="hide">Close</span>
</button>

more info here:
http://getbootstrap.com/javascript/#modals

Issue #448, opened 2015-12-02
Add security marking to listings and associated image files


blaino:
Every listing and every associated image file will need a valid CAPCO marking. We will eventually integrate with the Boostrap Classify library for handling the myriad possible classifications. This ticket is to pave the way for this eventual integration. In particular we need the following:

1. The create/edit page shall include as its first item a field for the security marking. 
2. Each image field (icons, banners and screenshots) shall include an additional field for its security marking. 
3. A modal, activated by a button, shall stand in for the Boostrap Classify modal, allowing the user to select a test classification level. 
4. The Quickview Details tab shall include the security marking. 
5. Listing Management's table view shall include a column with the security marking.

blaino:
@tonyalletag, @clarkjefcoat, @arw180, and @bhalt, please take a quick look to ensure I captured our discussion.

bhalt:
:+1:

tdressel:
Alan is still working on getting the CI box up and running.

tdressel:
1. - The create/edit page shall include as its first item a field for the security marking.  - PASSED.  Tested the security marking was there and working according to user's privs. Gave user friendly error upon save when high portion marking than what user was allowed.
2. - Each image field (icons, banners and screenshots) shall include an additional field for its security marking. PASSED- Jpg, Png and gifs using valid and invalid marking for each.
3. - A modal, activated by a button, shall stand in for the Boostrap Classify modal, allowing the user to select a test classification level. PASSED- Jpg, Png and gifs using valid and invalid marking for each.
4. - The Quickview Details tab shall include the security marking. PASSED shows accurate marking.
5. - Listing Management's table view shall include a column with the security marking.  - NOT TESTED - need a new cert in order to publish listing and see in table view.

tdressel:
Passed on Latest

tdressel:
These have been added.

tdressel:
Passed on staging.

PR #472, opened 2016-01-07
Feature security marking #448

blaino:
Requires the latest ozp-commons and a npm install (for commons and bootstrap-classify).


app/js/components/createEdit/validation/listing.js
```diff
@@ -59,6 +62,7 @@ function hasRequiredContactTypes (requiredContactTypes, contacts) {
 
 function ListingFull (requiredContactTypes) {
     return struct({
+        securityMarking: securityMarking,
```
wski: you can just do `securityMarking` with no `:` since we use ES6, this is more preference though so don't worry about changing it. Just wanted to let you know for the future if you so choose to.

package.json
```diff
@@ -77,6 +77,7 @@
   },
   "dependencies": {
     "icons": "git://github.com/ozone-development/ozp-icons",
+    "bootstrap-classify": "git://github.com/ozone-development/bootstrap-classify#ozp-integ",
```
wski: if we're forking this, perhaps master should be the ozp-integration

wski:
:+1:

clarkjefcoat:
Attempting to save an icon or an image without a security marking does not save.  Also choosing a Security Marking first, then Save, then choose an Image and Save results in a Bad Request.

I also found two bugs that do not appear to be related to this PR. (They also appear in the new-backend branch):
- Double-scrollbar in Chrome, Firefox on Ubuntu
- Associated Organization error

clarkjefcoat:
:+1:

wski:
:+1:

Issue #473, opened 2016-01-11
Search - Repeated Searches


clarkjefcoat:
Repeating a search after going to a different page does not work.

1.  Go main page type "test" into search box.
2.  From the hamburger menu choose "Listing Management"
3.  Return to the main page by clicking the shopping bag icon in the top menu bar.
4.  Type "test" into the search bar.

Expected Result:  Search results are displayed.

Actual Result:  The word "Searching..." is displayed on your screen forever (This is an estimate, I waited 30 seconds).

wski:
Can this be moved to resolved? @clarkjefcoat

tdressel:
Passed on Latest.  IE, FF and Chrome

tdressel:
Passed on new staging. IE, Chrome and FF

PR #474, opened 2016-01-11
fix(search): Allow consecutive searches

clarkjefcoat:
Issue #473


wski:
:+1:

blaino:
:+1:  Provided a reference to where debouncing is now handled.

clarkjefcoat:
This change assumes that all searches are deliberate.  And that even a search that is exactly the same as the previous search should be executed.

Back-to-back instances of the same search cannot be created inadvertently by a user due to the debounce method used in componentDidUpdate() of app/js/components/discovery/index.jsx.

PR #475, opened 2016-01-11
Remove bower

wski:


clarkjefcoat:
:+1:   So why don't we need the select2-browserify-fix anymore?

blaino:
SweetAlert and carouFredSel still work. :+1:

PR #476, opened 2016-01-12
Node upgrade

wski:
Needs to go in *AFTER* `remove-bower` @blaino @clarkjefcoat


blaino:
Downloaded and tested. Played around with Center for a few seconds - looked fine. Added a eslint error and it caught it. :+1:

clarkjefcoat:
:+1:

Issue #479, opened 2016-01-14
Double-scrollbar on Create Edit Page (Chrome, Firefox on Ubuntu)


blaino:
On the new-backend branch (CI-Prototype), there is double-scrollbar on the Create Edit Page when viewed in Chrome and Firefox on Ubuntu. Note that this does not happen in Chrome or Firefox on the Mac. It is currently unknown whether or not the problem exists on IE.

tdressel:
Passed on new Staging IE, FF and Chrome.

PR #480, opened 2016-01-15
fix(createEdit): Remove double scrollbar

clarkjefcoat:
Issue #479


blaino:
Pulled and tested. Works. But then again, on the Mac, I didn't have the problem in the first place. So: :+1:

wski:
:+1:

PR #497, opened 2016-02-01
fix(Dropdown menu)added aria-haspopup to SelectBox)

SOAP1914:
This is the Listing type and Organization Drop down (both should have focus)


SOAP1914:
@clarkjefcoat @wski  Two small changes (need to see if works in JAWS 17)

wski:
:+1:

blaino:
:+1:

PR #499, opened 2016-02-02
fix(Babel): Change parameter needed for IE10 support

clarkjefcoat:
For details see https://phabricator.babeljs.io/T3041


tonyalletag:
:+1:  - This has already been deployed to production via a hot fix

blaino:
Ok, then, :+1:

Issue #484, opened 2016-01-19
508 compliance - Add headings to center 


SOAP1914:
Headings provide structure to a page, A person using a screen reader can navigate a page quickly using headings within the application.

tdressel:
IE, CHROME and FF passed on Latest.

tdressel:
Insert + F6 displays heading list box and navigates accordingly.

PR #503, opened 2016-02-03
add(headings and focus)to the sidebar and model #484

SOAP1914:
add focus (tabIndex) and offscreen headings so JAWS can read when a user is looking for all headings within the app.


SOAP1914:
@blaino @arw180 @wski when you have time 2 look at it

blaino:
I pulled and clicked around. Doesn't seem to break anything. :+1:

SOAP1914:
Use an **off screen** technique to remove content from visible display, but still available to screen reader users:

wski:
looks good, In the future it would be cool to experiment with `display:none;` for the offscreen elements if the screen reader can read hidden elements. :+1:

SOAP1914:
I would love to use [display: none] but after doing some research this works for Center and JAWS 
 
https://www.paciellogroup.com/blog/2012/05/html5-accessibility-chops-hidden-and-aria-hidden/

PR #521, opened 2016-02-26
feat(IE < 10): stop users below IE10 from accessing AppsMall

wski:
Even though I don't think we should stop them from using AppsMall, this does it. Instead perhaps we should just do the warning, and let them continue. :warning: > :boom: @clarkjefcoat please review


clarkjefcoat:
Why does this PR include commits unrelated to the issue?

Please make the new window location configurable.  Maybe add it to OzoneConfig?

Issue #539, opened 2016-03-15
Remove redirects


clarkjefcoat:
The python backend is sending redirects (HTTP Status 301) when you connect to a URL without a trailing slash.  It redirects to the same URL with a trailing slash added.

Can/should we fix this by adding trailing slashing to URL requests?

clarkjefcoat:
The /category and /contact_type endpoints have this same problem.

clarkjefcoat:
Also /search

tonyalletag:
@skhtet @mannyrivera2010 Lets discuss this one

tdressel:
Passed on latest.

PR #540, opened 2016-03-15
fix(URL redirect):  Avoids browser redirects by adding trailing slash

clarkjefcoat:
#539


clarkjefcoat:
@wski - is this a reasonable thing to do?

wski:
Yeah looks :+1: will be incompatible with our old backend but I don't think that should matter.

tonyalletag:
:+1:

PR #552, opened 2016-03-20
chore(redirects): Conform to django standard of mandatory trailing slash

clarkjefcoat:
We should still discuss #539.

But for now we need to be consistent because the redirects are causing issues in our staging environment.


tonyalletag:
:+1:

PR #554, opened 2016-03-20
fix(package.json): Use a version of gulp-bless that uses bless 3.x.x

clarkjefcoat:
We are not compatible with the 3.1.x releases of gulp-bless, because that uses bless 4.  See
https://github.com/BlessCSS/gulp-bless/commits/master
https://github.com/BlessCSS/bless/issues/93


tonyalletag:
:+1:

PR #574, opened 2016-03-30
fix(Tour): Close tour at end of tour

wski:
@tonyalletag @clarkjefcoat


clarkjefcoat:
IE10 allows me to click 'Next' at the end of the tour.  In that case the modal is not closed.

wski:
I think that should be opened as a separate bug, the next button should not be clickable

clarkjefcoat:
:+1:

New Issue Created #575

akonsowski:
:+1:

Issue #419, opened 2015-10-22
What's New field does not allow 255 characters 


jal-vsi:
Populating a Listing's field "What's New" is limited to 250, not 255 as noted.  If you delete character 251 (last 'a' shown below) it saves properly.
![image](https://cloud.githubusercontent.com/assets/13243617/10668508/96621592-78aa-11e5-8b7b-65d337a70c0c.png)

clarkjefcoat:
The length of the what_is_new field in the database is 250.  So this is a bug with the message shown.

akonsowski:
419: fixed what is new issue#581[https://github.com/ozone-development/ozp-center/pull/581](url)

tdressel:
Passed on staging.

PR #581, opened 2016-04-11
changed whatisnew to 255 instead of 250

akonsowski:
#419 
@wski @clarkjefcoat @tonyalletag


wski:
:+1:

clarkjefcoat:
akonsowski and I confirmed that the database field will hold 255 characters.  This frontend change is enough to fix the problem.

:+1:

Issue #572, opened 2016-03-28
FIREFOX and CHrome ONLY- Thin black box displays around screenshots and rating descriptions


tdressel:
STR-
Go to an app with several screen shots
Open Pop Up Details
Scroll through the screenshots
RESULTS - An awkward thin black/blue box displays around the screen shots and description
EXPECTED RESULTS - Shouldn't display this thin black/blue box

Also happens when adding a rating description via popup details window.

tdressel:
<img width="897" alt="screen shot 2016-03-28 at 4 24 52 pm" src="https://cloud.githubusercontent.com/assets/5015370/14089243/f3b90aba-f501-11e5-8fe8-5a78c861b5e5.png">

rkenyon1969:
Matt - I've determined this can be resolved by adding a css attribute of outline:none to the * {} class in /dist/assets/main-blessed1.css. Could you walk me through how to modify/update that file? - thanks

@wski 
@tdressel 
@clarkjefcoat 
@akonsowski

wski:
This will fix it, however we should find out the root cause of our containers having this outline on `:active`. It's also important to note that adding the wildcard will remove it for elements where it should stay for accessibility reasons, such as input fields, textareas, buttons, etc.

rkenyon1969:
Agreed. My working theory is that the border is appearing because of the tabindex on the given element. In this case it is the <div> tag that make up the main sections of the "Overview" tab:

<div class="tab-pane active row quickview-overview" tabindex="0" data-reactid=".0.1.0.0.1.$tabs-container.1.0">

Question: (per Matt's point above) is this a 508 requirement or can we remove with either *{} or [tabindex]{} or on each individual element?

rkenyon1969:
@SOAP1914 - can you review to see if this needs a 508 review/input? - thanks!

rkenyon1969:
Ricardo has signed off that the border is not a 508 requirement and can be hidden. I will fix by applying above solution:

"I've determined this can be resolved by adding a css attribute of outline:none to the * {}"

rkenyon1969:
[tabindex] {
   outline: none !important;
}

[tabindex][input] {
   outline: 'default';
}

tdressel:
Passed on latest.

tdressel:
Passed on Staging.

PR #593, opened 2016-04-25
FIREFOX and CHrome ONLY- Thin black box displays around screenshots #572

rkenyon1969:
@clarkjefcoat 
@wski 
@SOAP1914

### DIFF
--- a/app/styles/main.scss
+++ b/app/styles/main.scss
@@ -53,6 +53,14 @@ strong {
   clear: both;
 }
 
+[tabindex] {
+   outline: none !important;
+}
+
+[tabindex][input] {
+   outline: 'default';
+}
+
 a:focus {
     outline-color: $blue;
 }
--- a/app/styles/main.scss
+++ b/app/styles/main.scss
@@ -53,25 +53,8 @@ strong {
   clear: both;
 }
 
-/*[tabindex] {
-   outline: none !important;
-   outline: 'default';
-}
-
-[tabindex][input] {
-   outline: none !important;
-}*/
-
-div:focus{
- outline: none !important;
-}
-
-section:focus{
- outline: none !important;
-}
-
-ul:focus{
- outline: none !important;
+[tabindex] {
+   outline: none;
 }
 
 a:focus {
--- a/app/styles/main.scss
+++ b/app/styles/main.scss
@@ -53,12 +53,25 @@ strong {
   clear: both;
 }
 
-[tabindex] {
+/*[tabindex] {
    outline: none !important;
+   outline: 'default';
 }
 
 [tabindex][input] {
-   outline: 'default';
+   outline: none !important;
+}*/
+
+div:focus{
+ outline: none !important;
+}
+
+section:focus{
+ outline: none !important;
+}
+
+ul:focus{
+ outline: none !important;
 }
 
 a:focus {


rkenyon1969:
thumbs anyone?

@clarkjefcoat @wski @SOAP1914

clarkjefcoat:
:+1:

SOAP1914:
👍 Mac's VoiceOver still works with this change

SOAP1914:
@rkenyon1969

app/styles/main.scss
```diff
@@ -53,6 +53,14 @@ strong {
   clear: both;
 }
 
+[tabindex] {
+   outline: none !important;
+}
```
wski: Not sure we want this one, there are things with a tabindex that we wan't highlighted. For instance the button in the notifications modal

wski:
👎 small change needed.

SOAP1914:
@rkenyon1969 @wski @clarkjefcoat  So I can still see the highlight in the top nav, which is what we want. But @rkenyon1969 was only touching the ones in popup modal Review tab (right?)

SOAP1914:
screen shot 2016-04-25 at 4 44 03 pm

wski:
This solution will work fine as long as we never add a tabindex to a form input

SOAP1914:
in the future we will need it for the form inputs @wski @rkenyon1969 @clarkjefcoat

wski:
nvm, i read it wrong 👍

wski:
nevermind x2, you will need to add an exception for `select` and `textarea` 👎

rkenyon1969:
Thank you all for your help. Let's try attempt part 2:

/*[tabindex] {
   outline: none !important;
   outline: 'default';
}

[tabindex][input] {
   outline: none !important;
}*/

div:focus{
 outline: none !important;
}

section:focus{
 outline: none !important;
}

ul:focus{
 outline: none !important;
}

bhalt:
I would stay away from using `!important` in our css as that quickly gets dangerous. I would also hesitate to override any form element behavior as it is currently set up in bootstrap/browser defaults -- since that adds both another layer of complexity to future development/compliance as well as more CSS to manage.

To solve this display issue, I think it would be as simple as adding 

[tabindex] { outline: 0; }

to the global CSS. This would remove the outline from anything that normally wouldn't have a focus outline but is inheriting one from the tabindex attribute (divs, etc) -- but leave the normal required behavior for form elements.

I haven't extensively tested this (just in Chrome) so it could use more testing to make sure it solves the issue. @clarkjefcoat

rkenyon1969:
@clarkjefcoat @SOAP1914 @wski @bhalt 

Question: Should we remove the outline? Is there a user/508 benefit leaving it alone. Browsers implement outline to allow users to visually determine where they are on a page.

Question: What if we standardized so that all outlines are the same; maybe with a unique(cool) design:

:focus {
 outline-style: dotted;
outline-color: #cc0000;
}

SOAP1914:
I'll have to see it first @rkenyon1969 (the unique design)it more of a nice to have. 
@clarkjefcoat @bhalt @wski what are your thoughts on this?

wski:
@bethany the reason he can't do just `[tabindex] { outline: 0; }` is be cause form elements need to have that outline, and some form inputs have `tabindex` thus there needs to be exceptions so when you're tabbing through a page, you know where the cursor is at.

wski:
halt

wski:
@rkenyon1969 making a standard outline might be more effort than it's worth because some browsers don't even allow you to remove it from things that have outline by default. (many mobile browsers and IE)

bhalt:
@wski When I added `[tabindex] { outline: 0; }` it didn't overwrite any form element focus displays. None of the current form elements in bootstrap get their outline from the tabindex attribute property, so that shouldn't change any form behavior, as long as the `!important` isn't added. Can you give more information if you tested it and found otherwise?

@rkenyon1969 we actually do customize the outline color in the OZP bootstrap but like Matt said, depending on the browser (and element that the element is on), the focus property will be interpreted different ways. If we make style changes, that should probably be a different PR than this bug fix.

rkenyon1969:
@clarkjefcoat @wski @bhalt @SOAP1914 

Ok. Final solution (please thumbs up):

[tabindex] {
   outline: none;
}

akonsowski:
:+1:

rkenyon1969:
@clarkjefcoat @wski @bhalt - I'd like to close this today if possible. Please thumbs up if you think it is ready. - thank you!

clarkjefcoat:
:+1:

Issue #561, opened 2016-03-22
Rejected State displays empty feedback or previous feedback not the most recent.


tdressel:
Found this on staging on all browsers.
STR-
Create a Listing
Submit a Listing
Listing Management
State | Pending
Double click on app tile
Go to Administration Tab
Reject the listing with a description
X out of the pop up modal
RESULTS - The App is now moved to Needs Action state
Go to Needs Action state
Choose drop down arrow in the right corner tile
Choose Read Feedback
RESULTS - If it's the first time being rejected it is empty.  If it has been rejected more than once it will show the previous time it was rejected. Click on All Listings and then back to My Listings and it now displays the most recent feedback.
EXPECTED RESULTS- It shouldn't require a refresh of page in order to display the most recent rejected feedback

tdressel:
Passed on Latest.

tdressel:
Passed on Staging.

PR #598, opened 2016-04-28
fix(myListings): users no longer have to refresh page to see latest r…

akonsowski:
…eject comments #561 added a fetch listings when changes occur to refresh the tiles with the latest comments 

@wski @clarkjefcoat @rkenyon1969 @SOAP1914

### DIFF
--- a/app/js/components/management/user/MyListings.jsx
+++ b/app/js/components/management/user/MyListings.jsx
@@ -116,6 +116,7 @@ var MyListings = React.createClass({
         this.setState({
             listings: this.getListings()
         });
+        ListingActions.fetchOwnedListings();
     },
 
     onFilterChanged: function (key, filter) {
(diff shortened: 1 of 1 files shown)


akonsowski:
please check it out

rkenyon1969:
Tested and solution works.

Q: what does the following code right above the new code do or not do?:

this.setState({
            listings: this.getListings()
        });

app/js/components/management/user/MyListings.jsx
```diff
@@ -116,6 +116,7 @@ var MyListings = React.createClass({
         this.setState({
             listings: this.getListings()
         });
+        ListingActions.fetchOwnedListings();
```
wski: even though I think this works, make sure you have thrown in a console.log here to make sure it's not infinity recursive if calling `ListingActions.fetchOwnedListings();` changes the store then it might be infinite.

akonsowski:
@wski you are right it was making an infinite loop. i have modified it to only update when the filter changes. let me know what you think.

wski:
should be good 👍

clarkjefcoat:
:+1:

rkenyon1969:
:+1:

Issue #583, opened 2016-04-19
Pop Up Details View - Org tag overlaps the App Title


tdressel:
STR-
Go to Center Settings | Organizations Tab | Add New - Title and Acronym called "DOD-Cats"
Add this Organization to an existing App ( with fairly long title) like "Location Viewer 9"
Save and reload the page.
Search on Location Viewer 9 
Drill into the app to the Pop up details view
RESULTS - Org tag overlaps the title
EXPECTED RESULTS - Should be properly spaced so it doesn't overlap. 

Screen shot attached.
<img width="900" alt="screen shot 2016-04-19 at 12 51 48 pm" src="https://cloud.githubusercontent.com/assets/5015370/14647980/15030512-062e-11e6-9be5-3887ac845fc1.png">

tdressel:
Passed on Latest.

tdressel:
Passed on Staging.

PR #599, opened 2016-04-29
fix(_quickviewscss): changed height and width and padding to make it …

akonsowski:
…fit the 8 character name #583 i tried to fix the padding so it look even.

@bhalt, @clarkjefcoat @wski @rkenyon1969

### DIFF
--- a/app/styles/_quickview.scss
+++ b/app/styles/_quickview.scss
@@ -118,13 +118,15 @@ ul.list-unstyled.RecentActivity
     padding: 25px;
     .company {
         background: $gray-lightest;
-        padding: 2px 10px;
+        padding: 2px 17px;
         border-radius: 10px;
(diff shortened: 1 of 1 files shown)


wski:
👍

rkenyon1969:
:+1:

bhalt:
@akonsowski Actually, it would be great to not have a height/width on this Org element, since that forces us to set a custom size every time we run into this issue again with a longer Org name. The element should have a dynamic width to flex with the content, rather than being boxed into certain pixel sizes.

To prevent word wrap when there is a dash, you can add 
```white-space: nowrap;```

to the `.company` and remove the other CSS added. Let me know if you need anything else!

akonsowski:
@bhalt that seems to have done the trick Thank you.

Issue #56, opened 2015-02-17
Create/Edit Page - Change body to 3 column style


mparizernc:
See mockups in OZF-678

sdeighan:
![mp_createeditform](https://cloud.githubusercontent.com/assets/7987446/6304853/4364d502-b8f5-11e4-8c34-8ef536b62ff8.png)

sdeighan:
UX Review completed. 3 column layout is working.

mparizernc:
Verified that the three columns appear correct in all the browsers.

PR #600, opened 2016-04-29
530 index

akonsowski:
this fix is to make the security marking when submitting a listing clearer to the user. please advise if i should change the wording to listing instead of application.  note  this fix requires pull request [#56](https://github.com/ozone-development/ozp-react-commons/pull/56) before working

### DIFF
--- a/app/js/components/createEdit/index.jsx
+++ b/app/js/components/createEdit/index.jsx
@@ -339,14 +339,14 @@ var ListingForm = React.createClass({
             durl.value = (durl.value) ? decodeURI(durl.value) : '';
             return durl;
         })();
-
+        
         return (
             <form ref="form" className="CreateEdit__form col-xs-9 col-lg-10">
                 <h2 id={f.basicInformation.id}>Basic Information</h2>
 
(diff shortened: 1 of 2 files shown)


app/js/components/createEdit/index.jsx
```diff
@@ -344,7 +344,7 @@ var ListingForm = React.createClass({
             <form ref="form" className="CreateEdit__form col-xs-9 col-lg-10">
                 <h2 id={f.basicInformation.id}>Basic Information</h2>
 
-                <MarkingInput id={f.securityMarking.id} { ...p('securityMarking') } aria-label="Classification and Control Marking"/>
+                <MarkingInput id={f.securityMarking.id} { ...p('basicSecurityMarking') } aria-label="Classification and Control Marking for the Application"/>
```
wski: What does ...p('basicSecurityMarking') return? We should just get the text for aria-label that way. I realize you didn't make the 'aria-label' but we should change how we pull the text for it if we're going to modify it anyway.

akonsowski:
@SOAP1914 can you check the aria-labels and make sure they are alight with the screen reader. i checked with the console log but i just want to make sure

SOAP1914:
👍

wski:
👍

PR #606, opened 2016-05-04
feat(Shrinkwrap): Brought back shrinkwrap to prevent futher issues with dependancy updates

wski:
@clarkjefcoat


clarkjefcoat:
:+1:

clarkjefcoat:
I think this will be an overall improvement.  But we will have to remember that we are using shrinkwrap when testing ozp-center with new versions of
- ozp-icons
- ozp-react-commons
- bootstrap-classify

akonsowski:
:+1: was able to build with shrinkwrap

Issue #608, opened 2016-05-06
Cant save a Listing


tdressel:
STR-
Submit a Listing
Save
RESULTS - Spinner appears with Uploading images.  Doesnt save
Save again
RESULTS - Nothing
EXPECTED RESULTS - Should save listing as a Draft and expose the publish button.

tdressel:
Passed on latest.

tdressel:
Passed on staging was able to save a listing.

PR #609, opened 2016-05-06
fix(index):previous security marking changes would require backend ch…

akonsowski:
previous security marking changes would require backend changes to save new listings changed text in security listing to make it save again
#608 
@clarkjefcoat @wski @mannyrivera2010


clarkjefcoat:
:+1:  Tested with https://github.com/ozone-development/ozp-react-commons/pull/58

wski:
:+1:

SOAP1914:
👍  (looked at it with @akonsowski )

Issue #529, opened 2016-03-11
Help Modal abruptly stops with overlay 


tdressel:
STR-
Go to AppsMall Center
Help icon
Take the Tour
RESULTS - Help Tour starts on HUD located on navigation bar
Advance through to the Tour until the pop up modal displays
RESULTS - FF stops the overlay immediately but user is able to navigate using the arrows but it eventually stops responding.  Chrome it stops loading the overlay for Overview Tab, then resumes on Reviews, Details, Resources and then it stops.
EXPECTED RESULTS - Should advance til the end.  At the end there should be something telling the user it is ending the tour.

clarkjefcoat:
This happens on FF 31.5

tdressel:
Both FF and Chrome advance properly and the NEXT button was removed from the last step in tour.

tdressel:
Passed on Staging.

PR #611, opened 2016-05-09
529 tour

akonsowski:
tour was skipping step 8. Also fixed the redirection bug that was causing the tour to stop in Firefox 31 and menu will now close when unfocused
#529 
@clarkjefcoat @mannyrivera2010 @SOAP1914 @wski


app/js/tour/centerTour.js
```diff
@@ -81,32 +89,42 @@ const meTour = new Tour({
...
+      onNext: function(){meTour.goTo(10)}
     },
+    //10
     {
+      path: `${CENTER_URL}#/home/?%2F%3F=`,
       element: ".Discovery__SearchResults .listing:first, .infiniteScroll .listing:first",
       title: "Listing Tiles",
       content: "Hover over a tile for a short description of the app, to bookmark it to your HUD, or to launch it into a new tab. Click the tile to open the listing overview for more listing details.",
       placement: "top",
+      onRedirectError: function(){ meTour.goTo(15) },
```
wski: `meTour.goTo(15)` needs `;`

make sure you're using eslint

app/js/tour/centerTour.js
```diff
@@ -246,6 +275,90 @@ const meTour = new Tour({
           }
         })();
       }
+    },
+    // end of tour the next following steps handle redirection errors
+    //15
```
wski: Maybe put a note here like 
```
/// HACK: Browsers that do not support the path key will automatically skip the above paths and jump to this step instead - @akonsowski
```

wski:
I've updated the package.json so that it doesn't conflict with the breaking changes we had. Please resolve merge conflicts though before merging.

wski:
Shrinkwrap still not changed :-1:

Issue #602, opened 2016-05-02
Add Listing DELETED to Activity Log and Listing Modal


mannyrivera2010:
This issue is for the front-end changes require for https://github.com/ozone-development/ozp-backend/issues/48

tdressel:
Displays as deleted in Listing Management | All Center Listings | List View
Displays as deleted in Listing Management | My Listings  Drill into pop up modal | Admin tab

tdressel:
Passed on staging.  When an app is deleted it is logged in the Recent Activity with username and on All Center Listings and My Listings as deleted.

PR #613, opened 2016-05-10
feat(components): added support for delete in listings

akonsowski:
updated center to work with new backend changed delete. the deleted listing will only be visible in all listings and recent activity once the listing is deleted it cant be recovered or modified see #602 

@clarkjefcoat @wski @SOAP1914 @rkenyon1969


SOAP1914:
👍  (looked at it with @akonsowski )

wski:
:+1:

Issue #617, opened 2016-05-12
Notification - See more sidebar displays wrong notification


dylankolson:
![screen shot 2016-05-12 at 11 27 22 am](https://cloud.githubusercontent.com/assets/10820108/15220280/e03665e0-1834-11e6-80ec-7197dd2dd422.png)
When you click a notification in the left sidebar it displays the other notification. Seems like the list on the left is reversed.

PR #618, opened 2016-05-12
(fix) 617 display wrong notification

dylankolson:
Fix for issue #617 where the see more notifications when clicked on displayed the wrong notification in the right of the pop up window.

### DIFF
--- a/app/js/components/NavBar/notificationsmodal.jsx
+++ b/app/js/components/NavBar/notificationsmodal.jsx
@@ -63,7 +63,7 @@ var NotificationsModal = React.createClass({
     makeSidebar: function() {
       var notis = this.state.notificationList.slice();
 
-      return notis.reverse().map((n, i) => {
+      return notis.map((n, i) => {
         var date = new Date(n.createdDate);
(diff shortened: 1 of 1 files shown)


wski:
Good catch, We will need to reverse the order further up. Currently the backend seems to be sending us notifications oldest first, and we want to get the reverse of that. In addition to removing the bad `reverse()` we should reverse the result in the webapi instead so that we see the newest notifications first. 

If you'd like I can tack this onto your PR

dylankolson:
@wski  Sure, the reason I removed it was I was working on adding a sort in ozp-backend for accessing the user notifications and it still displayed the wrong order on the sidebar but not in the navbar. 

https://github.com/dylankolson/ozp-backend/commit/be0abc117029024eaf352b82db012eefba791da2
https://github.com/ozone-development/ozp-backend/issues/160

wski:
Gotcha, if that's planned to go in soon I think this PR would be fine :+1: and we can hold off on adding anything to our webapi on center & hud until a sort is working in the backend. @mannyrivera2010 @clarkjefcoat

clarkjefcoat:
:+1:

Issue #625, opened 2016-05-18
notification module displays notification's date wrong


dylankolson:
Date is off by a month. For this example it should show May not June.

![screen shot 2016-05-18 at 11 31 12 am](https://cloud.githubusercontent.com/assets/10820108/15364897/2f69db4a-1cec-11e6-81fd-6aa528fa1ce3.png)

clarkjefcoat:
@dylankolson - Thanks for the bug report and the fix.  

But please don't close Issues for us.  We like for issues to remain open until they are closed by one of our testers.

tdressel:
Date displays accurately on latest.  Tested with expiration date in May and June and both worked.

PR #626, opened 2016-05-18
fix(notification): Display Correct Date #625

dylankolson:
Fix for displaying notification's date one month ahead.  date.getMonth() returns 0-11 which matches up with the array indexes thus +1 wasn't necessary
 
@clarkjefcoat 
@wski 
@akonsowski


akonsowski:
:+1:

wski:
Outlined in the ECMA Specification here: http://www.ecma-international.org/ecma-262/6.0/#sec-month-number Good catch @dylankolson :+1:

Issue #577, opened 2016-04-07
Listing Management - Recent Activity should show submitted listings


clarkjefcoat:

mannyrivera2010:
![image](https://cloud.githubusercontent.com/assets/1072762/14364014/e8df3b5c-fcd4-11e5-8637-0c977a2e1c5b.png)

It seems to be working

clarkjefcoat:
Try this as a normal user and not as an Admin.

akonsowski:
Issue moved to [ozone-development/ozp-backend #167](https://github.com/ozone-development/ozp-backend/issues/167) via [**ZenHub.io**](https://www.zenhub.io/)

akonsowski:
after making changes to the front end i am not getting expected results form the backend.
 received 
![image](https://cloud.githubusercontent.com/assets/18122042/15334224/4654f1f2-1c3b-11e6-8d58-bbd797074bb7.png) instead of 
![image](https://cloud.githubusercontent.com/assets/18122042/15334246/600e787a-1c3b-11e6-9d8d-8e8b77136594.png) 
@mannyrivera2010

tdressel:
Passed on latest.

tdressel:
Passed on staging

PR #627, opened 2016-05-18
fix(recentactivity): will allow the api to make the correct call to t…

akonsowski:
…eh back end for activity to the listings the user owns #577 

@clarkjefcoat 
@wski 
@SOAP1914 
@rkenyon1969


wski:
It appears that in the old way allowed us to look up other profiles changelogs. However this brings a regression in that we can only look up our own changelogs, is this intended? @clarkjefcoat

clarkjefcoat:
Yes, this is intended.  :+1:

PR #632, opened 2016-05-20
feat(Notifications): increase notification length to 600 characters

wski:
@clarkjefcoat @mannyrivera2010 @akonsowski


akonsowski:
:+1:  but it looks like this 
image

clarkjefcoat:
FYI, the database field that holds the notification message has a maximum length of 4096.  So we could accommodate messages even longer than the current 600 without changing the database if requirements change.

clarkjefcoat:
:+1:

wski:
@clarkjefcoat yeah I noticed that, I'm not sure we want to. @bhalt had mentioned she didn't want anything too high last time we spoke.

PR #635, opened 2016-05-25
fix(imageInput): Image input was not checking what file was being upl…

akonsowski:
…oaded. now if a user tries to upload an invalid image the webpage will give an error
@wski @rkenyon1969 @clarkjefcoat @SOAP1914

### DIFF
--- a/app/js/components/management/mall/notifications/CreateNotification.jsx
+++ b/app/js/components/management/mall/notifications/CreateNotification.jsx
@@ -57,7 +57,7 @@ var CreateNotification = React.createClass({
 
     onMessageChange(event) {
         var { value } = event.target;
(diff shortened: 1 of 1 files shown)


wski:
Please rebase these together once the unit tests pass

app/js/components/createEdit/form/ImageInput.jsx
```diff
@@ -69,6 +69,7 @@ var ImageInput = React.createClass({
         if(value.type === 'image/gif' || value.type === 'image/png' || value.type === 'image/jpeg'){
           if ( value.size <= 1024000){
             this.props.setter(value);
+            console.log(value)
```
wski: This will also fail the unit tests, shouldn't be checked in

Issue #546, opened 2016-03-16
Listing - Viewing Reviews for a Listing creates two duplicate calls to backend  


mannyrivera2010:
Viewing Reviews for a Listing creates two duplicate calls to Backend  

![image](https://cloud.githubusercontent.com/assets/1072762/13828086/1fa3ab82-eb96-11e5-914c-218735a224cc.png)

wski:
@akonsowski is this what you were working on the other day?

wski:
So the reason this happens is because of the component will receive props, in the case that you're not transitioning to another application it hits the backend twice, but this should only happen the first time.

wski:
On second thought I believe the issue is that the reviews tab is calling componentWillMount twice, so somehow we're mounting two reviews tabs.

akonsowski:
it looks like the will receive props and will mount are getting called at the same time.

tdressel:
Test with Bug fix 656

tdressel:
Test on staging and only saw one call.  Passed.

PR #640, opened 2016-05-31
fix(Index): componenet will mount and will receive props were both ru…

akonsowski:
…nning when the app loads the review. the reviews will still load when will receive prrops is called removed will load #546


akonsowski:
@wski @SOAP1914 @rkenyon1969

wski:
I'll need to take a closer look at this, judging by the comment if there is no review found in the store it will fail.

wski:
:+1: but I think @tdressel should do some testing on this before we merge

rkenyon1969:
:+1:

PR #645, opened 2016-06-07
fix(Search)Made search result tab to main content

SOAP1914:
@wski @akonsowski @rkenyon1969


akonsowski:
:-1: when focus changes and goes to the listings a lot of things can happen. like going back a page if you hit backspace, or typing with no effect in the search bar. i would suggest changing focus after hitting enter or something like that

SOAP1914:
Thanks @akonsowski for the bug catch.

wski:
:+1:

akonsowski:
:+1:

Issue #646, opened 2016-06-13
Listing Mgmt | All Center Listings | Deleted (state) - undefinedundefinedundefined


tdressel:
STR-
Log onto Apps Mall 
Listing Management
All Center Listings
Search on a known App
After results display, Choose Deleted (State) filter on the left
RESULTS- undefinedundefinedundefined
EXPECTED RESULTS - Should display results with filter or delete the filter when the State is changed.

tdressel:
Passed on latest

tdressel:
Passed on staging.

PR #648, opened 2016-06-20
646 alllistings

akonsowski:
#646 
it looks like app/js/components/management/AllListings.jsx was  part of the old way on how center listings were called and displayed this jsx is never called in the project. I also changed the functionalityof changing filters to clear search parameters  when changed.
@wski 
@rkenyon1969 
@clarkjefcoat 
@SOAP1914


wski:
:+1: for reset, scared about nuking an entire file. @tdressel could you run through our use cases just to be sure?

rkenyon1969:
:+1: - same concerns as Matt. Plus not versed enough to sigh off on w2ui change.

Issue #653, opened 2016-06-23
add a stop for notifications in tour


akonsowski:
we need to add a stop in tour for the notifications tab

PR #654, opened 2016-06-23
feat(centerTour): added new stop for notification tab

akonsowski:
#653 
@wski @rkenyon1969 @clarkjefcoat 
@bhalt any suggestion for the wording about what the tab is about?


bhalt:
@akonsowski As far as I know, we don't currently have a tour stop for the Administration tab. The Notifications tab is also an admin tab. Not sure of the capabilities to show a tour stop depending on the user role? This also gets tricky because even if a user has the creator role, if the first detailed view that comes up in the tour isn't one that they created, they won't see the tabs. This may be a later feature? @clarkjefcoat 

If we do have capabilities to add this now, language could be:

**Notifications**
As the creator of a listing, you can send notifications to users who have bookmarked your listing.

akonsowski:
@bhalt thank you for the clarification. i was under the impression you could check out notifications in the tab as a regular user. i will close this PR

Issue #614, opened 2016-05-11
FF Only - Help Tour Previous step loops back incorrectly


tdressel:
STr-
Go to Help tour
Navigate to Center Home , Listing Tiles then Listing Overview 
Choose Previous twice
RESULTS - Loops back to the Listing Overview
EXPECTED RESULTS - Should Navigate back to Center Home

tdressel:
Passed on Latest.

tdressel:
Passed on Staging.

PR #657, opened 2016-06-24
fix(centerTour):fixed tour bug

akonsowski:
#614 
@wski @rkenyon1969 @clarkjefcoat


wski:
Duplicating the entire tour seems like a very bad idea for long term maintainability. Let's investigate this further and see if we can't fork the bootstrap tour library and fix the error all together.

PR #662, opened 2016-06-29
614 ffbug

akonsowski:
#614 

i removed global tour as it was not used and also removed redirection errors as they were fixed in the bootstraptour.js itself.

@wski @clarkjefcoat @rkenyon1969


clarkjefcoat:
:+1:

rkenyon1969:
:+1:

PR #684, opened 2016-08-03
fix(index):admin tab will not display for non admin users and will al…

akonsowski:
…so not display a hyperlink for owners in details tab

### DIFF
--- a/app/js/components/quickview/DetailsTab.jsx
+++ b/app/js/components/quickview/DetailsTab.jsx
@@ -115,18 +115,18 @@ var DetailsTab = React.createClass({
 
     renderOwners: function () {
         var owners = this.props.listing.owners;
-        if(!owners){
-        return owners.map(function (owner, i) {
-            return (
-                <p className="listing-owner" key={`renderOwners.${i}`}>
-                    <span> </span>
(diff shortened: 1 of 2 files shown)


akonsowski:
@wski @clarkjefcoat @rkenyon1969

wski:
:+1:

clarkjefcoat:
Make changes as discussed before accepting.

wski:
@akonsowski can you verify I've added everything you had locally?

clarkjefcoat:
I still don't understand what this code does.  Please test all of the following scenarios before accepting:

A normal user with the anonymize function set to return True and the user is not the owner
An admin user with the anonymize function set to return True and the user is not the owner
A normal user with the anonymize function set to return False and the user is not the owner
An admin user with the anonymize function set to return False and the user is not the owner
A normal user with the anonymize function set to return True and the user is the owner
An admin user with the anonymize function set to return True and the user is the owner
A normal user with the anonymize function set to return False and the user is the owner
An admin user with the anonymize function set to return False and the user is the owner

akonsowski:
@skhtet

skhtet:
:+1:

rkenyon1969:
:+1:

Issue #693, opened 2016-08-18
2PI - QuickView Details | Reviews Tab | edit icon is available to user


tdressel:
STR-
Go to QuickView Details | Reviews tab 
Choose Edit icon for the review
RESULTS - Edit Review form is available for user
EXPECTED RESULTS- Edit Review icon should not be available to user

*****it does not respond to Submit button so it technically doesn't allow you to edit anyway******

tdressel:
Edit icon only displays for current user's review.
Edit icon displays for all reviews when logged in as admin.

tdressel:
Log on as current user
Made a review
Refreshed
Edit icon was available to me but my own user name was masked.

tdressel:
ested and works with 2pI user. Needs testing with 1st party user.

tdressel:
Passed on Staging.

Issue #692, opened 2016-08-18
2PI - QuickView Details | Reviews Tab | Allows users to delete reviews


tdressel:
STR-
Go to QuickView Details | Reviews Tab 
Delete an existing review
RESULTS - Review is deleted
EXPECTED RESULTS - 2PI user should never be able to delete/edit someone else's review. They should be able to edit and delete their own.

tdressel:
Tested on Latest.  
All users (  Basic users/app owners, admins and org stewards) can delete their own reviews successfully.
Admins can delete other's reviews successfully.
Non admins cannot delete reviews that are from someone else.

tdressel:
2PI Users passed but needs testing with a 1st Party admin

tdressel:
Admins and orginal owners can delete reviews but 2PI and any others that are not original owners can not.

Issue #679, opened 2016-07-28
Pop Up Details | REVIEW TAB | Edit icon is available to other users 


tdressel:
STR-
Log on as basic user
Go to App with reviews (the reviews have to be made by someone else)
RESULTS- Edit icon is available.
EXPECTED RESULTS - Edit icon shouldn't be available to anyone but the reviewer.

The problem with this is that the user can edit the review and when they submit it .......it doesn't save the edited version.

tdressel:
Edit icon only displays for current user's review.
Edit icon displays for all reviews when logged in as admin.

akonsowski:
admins should be able to view other users reviews only for the purpose to delete them

tdressel:
Passed on Latest.  Test admin, reviewer and non reviewer.

tdressel:
Tested and works with 2pI user.  Needs testing with 1st party user.

tdressel:
Passed on Staging.

PR #695, opened 2016-08-22
fix(reviews):removed edit icon for masked users

akonsowski:
#693 #692 #679 
only admins/ org stewards should be able to delete reviews from anyone review authors should only be able to delete / edit their own reviews and 2pi users will only be able to submit a review and not be able to edit it.
@wski @skhtet @rkenyon1969


wski:
:+1:

rkenyon1969:
:+1:

PR #696, opened 2016-08-23
fix(package):fs package is now part of the npm core

akonsowski:
having the fs package install was breaking travis build
[source](https://github.com/npm/npm/issues/13743)


akonsowski:
@wski @rkenyon1969 @skhtet

wski:
:+1:

skhtet:
:+1:

wski:
Agreed that we should not have been requiring built in modules from the start. Go ahead and merge IMO.

PR #702, opened 2016-08-26
fix(editreview):disabled review editing except for own reviews

akonsowski:
#679 
when viewing a review that is not the current users the edit box will be deactavited
@wski @skhtet @rkenyon1969


wski:
:-1: :-1: :-1: :-1: :-1: :-1: :-1: :-1:

skhtet:
:+1:

Issue #706, opened 2016-08-31
saving a listing with 2pi user fails


akonsowski:
when saving a listing trying to add a 2pi user as an owner the listing will  fail and get stuck

tdressel:
Passed on latest.  Validation stopped an org steward from adding a 2PI user to an existing app.

tdressel:
Passed on 2PI environment

tdressel:
Passed on Staging

PR #707, opened 2016-08-31
fix(index):added a check to see if the backend is throwing additional…

akonsowski:
… error codes that are not checked for in the validation.
reference to #706 
@wski @rkenyon1969 @blynch11p


rkenyon1969:
:+1:

mannyrivera2010:
:+1:

PR #717, opened 2016-09-15
fix(index):added a check to see if we should display webtop

akonsowski:


akonsowski:
@blynch11p @wski @rkenyon1969

blynch11p:
:+1:

wski:
:-1:

Issue #524, opened 2016-03-08
Tabbing through the carousel causes the previous/next buttons to move awkwardly in front of apps and out of alignment


tdressel:
Tabbing through the carousel causes the previous/next buttons to move awkwardly in front of apps and out of alignment

wski:
Can you find a way to reproduce this LS? If not it's going to be impossible for most of us to address.

lsemesky:
> Can you find a way to reproduce this LS? If not it's going to be impossible for most of us to address.

Seconded.

lsemesky:
After discussion with Tiff there is no way for me actually work on this issue since it is HS only. Moving to backlog.

blynch11p:
This is reproducible on LS. I showed @akonsowski this morning.
Most of the time, the Arrows/buttons just disappear from the carousel, but sometimes they shift away from their correct positioning.

tdressel:
I was able to recreate in FF34.

tdressel:
Passed on Staging. IE10, FF 31 and Chrome 36

PR #721, opened 2016-10-03
fix(Carousel): Fixed arrows disappearing when tabbing into carousel #524

lsemesky:
Fixed the arrow buttons disappearing by moving them into the parent container so that they did not move when tabbing @akonsowski @blynch11p


akonsowski:
looks good to me :+1:

blynch11p:
:+1:

Issue #709, opened 2016-09-02
2PI- Security Banner reflecting incorrectly for 2PI Users


tdressel:
See Soe or Tiffany for explanation.

tdressel:
Apps Mall should reflect the highest security banner of current user not of highest security marking in app.

lsemesky:
@skhtet will need to look at the logic for "Apps Mall should reflect the highest security banner of current user not of highest security marking in app." 

Best I could do was update it to show a different banner for second party users.

tdressel:
Needs testing on Staging.

tdressel:
Confirmed working for 2PI user on Staging.

PR #724, opened 2016-10-11
Fix(SecurityBanner): display different banner for second party users

lsemesky:
Fix for issue #709  @skhtet @blynch11p @rkenyon1969


blynch11p:
:+1: 
Looks good, functionally appears to work per different user logins.

skhtet:
:+1:
Works on dev, ops problem now.

Issue #723, opened 2016-10-07
New Arrivals carousel wont scroll.


tdressel:
STR-
Choose Right arrow on New Arrivals to scroll over 
RESULTS - it opens the app listing it is overlapping

tdressel:
Passed in Latest

tdressel:
Passed on Staging. IE10, FF 31 and Chrome 36

PR #725, opened 2016-10-11
Fix(Carousel): Moved carousel buttons below carousel div to correct c…

lsemesky:
…lickthrough #723 

@blynch11p @skhtet @rkenyon1969


blynch11p:
:+1:

PR #741, opened 2016-11-21
fix(details):tags have been updates to work on all browser versions

akonsowski:
@lsemesky @blynch11p @rkenyon1969 @mannyrivera2010 
the routes were not working because navigation was on home page when viewing the model


blynch11p:
This is really just for internal linking, right?  Not opening apps externally?  If #1, then I think we're ok.  It's not a metric that's tracked now (though, it could be)

blynch11p:
Looks good to me
:+1:

lsemesky:
:+1:

PR #746, opened 2016-12-05
enhance(typeahead): Application details now popup when application na…

lsemesky:
…me is selected in typeahead AMLNG-114
Currently when an option in the type down is selected it just replaces the search term and does another search which now thanks to smart search will return more than the exact name of the application. Instead the desired behavior is when the name of the app is selected in the drop down it will also open the application as if the search result was single clicked
Acceptance criteria:
Type in a search result into the search bar
Select one of the applications listed in the dropdown
The application preview is opened as if it were single clicked from the search result page.
Other existing behavior is not changed


lsemesky:
@akonsowski @blynch11p @rkenyon1969

lsemesky:
Note: Need to pull backend master branch for this to work right

blynch11p:
Looks good :+1: 

Unrelated to the checkin, but related to the suggestion list...
Is/was there a style for the suggestion list for mouseover or arrow up/down to highlight what's selected?  Both of the dropsdown to the right show a hover indicator. I'd expect this to be similar.

akonsowski:
:+1:

PR #749, opened 2016-12-13
fix(typeahead): fixed issue with typeahead not correctly inputting AML-185

lsemesky:
@blynch11p 
@akonsowski 
@skhtet


J-Fid:
:+1:

akonsowski:
:+1:

blynch11p:
Sorry, disagree :-1:

blynch11p:
git isn't letting me push up my changes, so...
ozp/center/app/js/utils/typeahead.js
~line 686: 
`this.limit = o.limit || 10;`
and
~line 823
`suggestion = (suggestion || []).slice(0, that.limit);`
`rendered = suggestion.length;`
`that._overwrite(query, suggestions);`

lsemesky:
@blynch11p  I agree that the typeahead util had something wonky with the logic but didn't think it was appropriate to change a third party library. I figured reverting to the previous appsmall 5 suggestion limit was acceptable and better than introducing a breaking point in case of upgrading the typeahead library. Your thoughts?

lsemesky:
If we want to up the limit at a later point we could probably pass in the limit and make sure to use async and it should be ok

blynch11p:
I'll retract my statement.  Still seems to be something no right with the 3rd party library, but the fix works, as long as we're ok with the 5 limit and not 10 limit.

PR #752, opened 2017-01-05
Adding a style specifically for FF Browse button on Listing Managemen…

blynch11p:
…t page

@lsemesky @akonsowski @rkenyon1969 

This works, updates the style for FF only, does not affect Chrome.  Welcome to suggestions if there's a different way any of you would approach it however.


akonsowski:
:+1:

PR #753, opened 2017-01-09
Re-init-ing Notifications after a new fetch.

blynch11p:
@lsemesky @akonsowski @rkenyon1969 
The notifications were being built on top of each other over and over causing non-unique keys, this re-inits the notifications after a completed fetch.


akonsowski:
:+1:

PR #757, opened 2017-01-26
added a check to verify the text in the search box was actually the d…

akonsowski:
…esired search term
@blynch11p @lsemesky

### DIFF
--- a/app/js/components/management/shared/TableView.jsx
+++ b/app/js/components/management/shared/TableView.jsx
@@ -98,10 +98,7 @@ var TableView = React.createClass({
 
                 thisTable.props.filter.offset = 0;
                 thisTable.props.filter.search = event.searchValue;
-                if(event.searchData)
-                  if(event.searchData[0].value === '')
-                    delete thisTable.props.filter.search;
-                if(!event.searchValue)
+                if(!event.searchValue || !event.searchData[0])
                     delete thisTable.props.filter.search;
                 UnpaginatedListingsStore.filterChange(thisTable.props.filter);
             },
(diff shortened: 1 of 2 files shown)


app/js/components/management/shared/TableView.jsx
```diff
@@ -74,36 +74,39 @@ var TableView = React.createClass({
...
                 }
                 thisTable.props.filter.ordering = thisTable.convertToServerPropName(event.field, ascending);
                 UnpaginatedListingsStore.filterChange(thisTable.props.filter);
             },
             onSearch: function (event){
-                
+
                 thisTable.props.filter.offset = 0;
                 thisTable.props.filter.search = event.searchValue;
+                if(event.searchData)
```
lsemesky: this code needs to be changed and can be consolidated to `if(!event.searchData.length || !event.searchData[0].data)`  or `if(!(event.searchData.length && event.searchData[0].data)`

this will remove the need for the `event.searchValue` check since when searchValue is undefined, grid also has `event.searchData.length  = 0` (in current code your `event.searchData` returns true when the x is clicked, but `event.searchData[0]` is undefined and you can't access value of undefined as it currently is)

empty strings in javascript evaluate as falsey (as well as if they are undefined or null) so instead of checking 
`event.searchData[0].value === ''` we can just check `event.searchData[0]` and if it is null, undefined or empty it will return false

lsemesky:
:+1:

blynch11p:
Ignoring two somewhat minor things :+1: 

- The (X) is still there when clearing out the text field and hitting enter.  It doesn't affect functionality or search results. 
- I get a 'ReferenceError: event is not defined' error when clicking in the text box, but I think @akonsowski showed me that he does not.

akonsowski:
:+1:

PR #758, opened 2017-01-27
(fix):when undeleting the listing the is enables status is set to tru…

akonsowski:
…e instead of keeping it disabled
 @skhtet @blynch11p @lsemesky


blynch11p:
:+1:

lsemesky:
:+1:

PR #767, opened 2017-03-13
fixed pending deletion alignment

lsemesky:
@blynch11p @mleeBoeing in order to give a thumbs up for this you need to checkout this branch and make sure it works according to the following details:

  
image

Other categories word wrap in line with the first word when browser is resized. May need to add a missing css class somewhere to conform

Then either :+1:  (+1 inside of colons) or leave comments on the code

### DIFF
--- a/app/styles/management/user/_my-listings.scss
+++ b/app/styles/management/user/_my-listings.scss
@@ -113,7 +113,7 @@
         }
     }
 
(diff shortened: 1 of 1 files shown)


blynch11p:
:+1:

J-Fid:
:+1:

PR #774, opened 2017-03-21
JIRA-337: Bug fix for drop down letters being cut off

mleeBoeing:
Bug fix for JIRA-337. Drop down letters were being cut off


blynch11p:
:+1:

PR #779, opened 2017-03-29
508

mleeBoeing:
Notification should describe how many results are displaying on the field and now talks about # of results.

### DIFF
--- a/app/js/components/discovery/index.jsx
+++ b/app/js/components/discovery/index.jsx
@@ -466,6 +466,9 @@ var Discovery = React.createClass({
                     &nbsp;<span className="icon-share-10-blueDark"></span>
                   </span>
                 </h4>
+                <div>
+                    Showing {this.state.searchResults.length} out of {DiscoveryPageStore.getTotalSearchResults()} listings.
+                </div>
                 <p><DetailedQuery
                   onCategoryChange={this.onCategoryChange}
                   onTypeChange={this.onTypeChange}
--- a/app/js/components/discovery/index.jsx
+++ b/app/js/components/discovery/index.jsx
@@ -455,11 +455,9 @@ var Discovery = React.createClass({
         }
 
         if (this.state.searchResults.length != 0) {
-            listingResults = <div aira-label={'Showing (this.state.searchResults.length) out of (DiscoveryPageStore.getTotalSearchResults()) listings'}>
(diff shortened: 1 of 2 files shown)


app/js/components/discovery/index.jsx
```diff
@@ -446,13 +446,22 @@ var Discovery = React.createClass({
...
+        let listingResults;
 
         if (!this._searching) {
             results = this.state.searchResults.length > 0 ?
                 ListingTile.fromArray(this.state.searchResults) :
                 <h3 className="col-xs-12">No results found.</h3>;
         }
 
+        if (this.state.searchResults.length != 0) {
+            listingResults = <div aira-label={'Showing (this.state.searchResults.length) out of (DiscoveryPageStore.getTotalSearchResults()) listings'}>
```
blynch11p: aria, not aira.

blynch11p: The variables (this.state.searchResults.length) are showing exactly as typed in the source after fixing the typo.  You might have to build the text elsewhere, as a variable, to put in the html.

lsemesky: try using the jsx syntax by surrounding variables (or computed statements) in brakets like 'Showing {this.state.searchResults.length} ....'  (I think that will work- play around with it if not)

app/js/components/discovery/index.jsx
```diff
@@ -446,13 +446,22 @@ var Discovery = React.createClass({
...
+                Showing {this.state.searchResults.length} out of {DiscoveryPageStore.getTotalSearchResults()} listings.
+            </div>
+        } else {
+            listingResults = null;
```
lsemesky: if you initialize listingResults to null you don't need the else

blynch11p:
:+1:

blynch11p:
:+1:

app/js/components/discovery/index.jsx
```diff
@@ -446,13 +446,22 @@ var Discovery = React.createClass({
...
         var results = '';
+        let listingResults;
 
         if (!this._searching) {
             results = this.state.searchResults.length > 0 ?
                 ListingTile.fromArray(this.state.searchResults) :
                 <h3 className="col-xs-12">No results found.</h3>;
         }
 
+        if (this.state.searchResults.length != 0) {
```
mleeBoeing: :+1:. Will use this from now on

PR #786, opened 2017-04-13
fixed typo that mlee identified

lsemesky:
Separating this from mlee's pull request so this can be corrected before his pr is merged

@mleeBoeing @blynch11p


blynch11p:
:+1:

PR #789, opened 2017-04-19
 Changelogs limit

blynch11p:
@mleeBoeing @J-Fid @lsemesky 

Removed the unused code in Listing.js.  
Prior PR wasn't updating the Listing Changes when moving between different Application quickview displays.  This should fix that.


mleeBoeing:
:+1:

PR #794, opened 2017-04-27
JIRA 428 - Similar style fix

lsemesky:
image

when an app had a really long title. it was wrapping the text and overflowing like above. Now it should truncate with elipses like:

image

@blynch11p @mleeBoeing


blynch11p:
:+1:

PR #795, opened 2017-05-01
fix(notification): notification modal no longer says "0No Notifications" JIRA-402

lsemesky:
@blynch11p @mleeBoeing 

image

To get to that screen as seen above:

make sure you have one notification, click on bell, click on the "See More" button, dismiss all notifications... the message should be similar to the above image just with no 0 in front


blynch11p:
:+1:

PR #796, opened 2017-05-08
initial commit for screenshot re-arrangement

mleeBoeing:


blynch11p:
w00t
:+1:

app/js/components/createEdit/form/ListInput.jsx
```diff
@@ -53,9 +56,34 @@ var ListInput = React.createClass({
         event.preventDefault();
         var items = this.props.value;
         items.splice(key, 1);
+        this.props.setter(null);
```
lsemesky: setting breakpoints I learned it is changing the property that corresponds to the section (in this case screenshots). My best guess is that since this form isn't listening to the store anywhere and there is no custom function for componentshouldupdate that setting the prop to null first makes the componentshouldupdate true and so the component updates. That's just my best guess though.

lsemesky:
:+1:

PR #802, opened 2017-05-22
Fix hanging issue when submitting new listing

mleeBoeing:
Fix for when you try to submit listing and it hangs on the saving listing popup. 

@lsemesky - I have done some tests to make sure this push won't break similar listings. However, please confirm.

### DIFF
--- a/app/js/stores/CurrentListingStore.js
+++ b/app/js/stores/CurrentListingStore.js
@@ -133,6 +133,7 @@ var CurrentListingStore = createStore({
     refreshListing: function (listing) {
         revokeAllObjectURLs();
         _listing = listing;
+        _listingId = listing.id;
         _submitting = false;
         var validation = this.doValidation();
         this.trigger({


blynch11p:
:+1: 
Everything seems to be working

app/js/stores/CurrentListingStore.js
```diff
@@ -157,6 +157,14 @@ var CurrentListingStore = createStore({
             if(newListing){
                 this.refreshListing(newListing);
             }
+        } 
+        // Bug fix for hanging submit listings. Code from commit fedf28b
```
lsemesky: instead of this, you can add a line in refreshListing function (line 133) to say `_listingId = listing.id;` which seems to fix it. I believe that was why the onCacheUpdated wasn't working right because the new _listingId never got set once the globalcache was updated after the initial save

rkenyon1969:
:+1: 

- Reviewed code
- discussed logic with @mleeBoeing
- successfully tested branch locally

PR #806, opened 2017-06-05
fix: Selecting the center icon now takes the user to the Center Homepage.

J-Fid:
Please Review:
@lsemesky 
@mleeBoeing 
@mannyrivera2010 
@blynch11p 

To test, enter something in the search box.  Afterwards, select the center icon to return to the center homepage.

### DIFF
--- a/app/js/components/NavBar/index.jsx
+++ b/app/js/components/NavBar/index.jsx
@@ -50,7 +50,7 @@ var NavBar = React.createClass({
                     <div className="navbar-left">
                         <ul className="nav navbar-nav">
                             <li id="tourstop-hud" className="tooltiped" data-toggle="tooltip" data-placement="bottom" title="HUD"><a className="lrg" href={HUD_URL}><i className="icon-home-grayLightest"></i><span className="hidden-span">HUD</span></a></li>
-                            <li id="tourstop-center" className="active tooltiped" data-toggle="tooltip" data-placement="bottom" title="Center"><a className="lrg" href='#/home' ><i className="icon-shopping-grayLightest activeIcon"></i><span className="hidden-span">Center</span></a></li>
+                            <li id="tourstop-center" className="active tooltiped" data-toggle="tooltip" data-placement="bottom" title="Center"><a className="lrg" href='' ><i className="icon-shopping-grayLightest activeIcon"></i><span className="hidden-span">Center</span></a></li>
                             { !secondParty &&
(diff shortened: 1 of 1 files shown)


lsemesky:
This technically works but I don't like the solution because it forces an entire refresh of the page which causes all the info to need to be downloaded again instead of making use of React's caching. Ideally this link would work the same as clicking the "Center Home" in the sidebar. (Take a look at /app/js/components/discovery/Sidebar) 

If others think it's not worth the effort of refactoring this (as it would require more knowledge of react), I would consider at least making it consistent with the other header links by changing line 9 from 
```var { HUD_URL, METRICS_URL, WEBTOP_URL, DEVELOPER_RESOURCES_URL} = require('ozp-react-commons/OzoneConfig');``` 
to 
```var { HUD_URL, METRICS_URL, WEBTOP_URL, DEVELOPER_RESOURCES_URL, CENTER_URL} = require('ozp-react-commons/OzoneConfig');```

and change ```href=''``` to ```href={CENTER_URL}```

lsemesky:
If @blynch11p or @rkenyon1969 decides it's fine as-is I am fine thumbing up as it works fine

lsemesky:
Just read the Jira ticket and noticed it does have acceptance requirement as this icon doing the same thing as clicking the logo, which it does. I also believe the behavior of the logo should change as well, but perhaps that's for a different tech debt ticket.

blynch11p:
Laura makes good points.  What this ticket brings up is the fact the the URL is not correctly updated when filtering the home page by Organization, Listing Type, or Category.  The routing for those all work (there is a `Share` link available after running any of these filters that can be copy/pasted and the correct page is built).  Though, clicking the shopping bag after using that link does not correctly take the user back to the Center home even though the URL updates.

I think I may just lean toward this fix being OK.  The functionality is consistent with the Logo, and all of the links across the top (HUD/Webtop) are also 'external links' that cause the full page to load.

I'm a :+1: unless @rkenyon1969 has a different weigh-in.

rkenyon1969:
Please let me review before proceeding. I will take a look after stand-up. - thank you,

rkenyon1969:
I agree with Laura and Brian - Fixes current issue and we can discuss future enhancements later.

- Pulled the branch
- Reviewed the code
- Tested successfully

:+1:

rkenyon1969:
Spoke w/ Brian. Need to create additional ticket for backlog to address remaining issue (see below):

- Pulled the branch
- Reviewed the code
- Tested successfully

:+1: 


**Remaining Issues:**
- Listing API called twice
- Do we need if(!_listing)
- What is "new Listing({ owners: [this.currentUser] })" for?

PR #807, opened 2017-06-06
Initial commit for tag subscription

mleeBoeing:
Need to pull backend branch as well: https://github.com/aml-development/ozp-backend/pull/309

### DIFF
--- a/app/js/components/discovery/DetailedQuery.jsx
+++ b/app/js/components/discovery/DetailedQuery.jsx
@@ -192,7 +192,12 @@ var DetailedQuery = React.createClass({
     },
 
     render() {
-        var subscribeLink = <a className="subscribe" onClick={(e) => {this.subscribeToTag(e)}} >Subscribe</a>;
+        if (this.props.data.tagId.length > 0) {
+            var subscribeLink = <a className="subscribe" onClick={(e) => {this.subscribeToTag(e)}} >Subscribe</a>;
+        } else {
+            subscribeLink = null;
+        }
+
         if (this.state.subscriptionStore && this.state.subscriptionStore.length > 0) {
             for (var element of this.state.subscriptionStore) {
                 if (element.entity_id == this.props.data.tagId && element.entity_type === "tag") {
--- a/app/js/components/discovery/DetailedQuery.jsx
+++ b/app/js/components/discovery/DetailedQuery.jsx
@@ -192,17 +192,18 @@ var DetailedQuery = React.createClass({
     },
 
     render() {
-        if (this.props.data.tagId.length > 0) {
-            var subscribeLink = <a className="tag_subscribe" onClick={(e) => {this.subscribeToTag(e)}} >Subscribe</a>;
-        } else {
-            subscribeLink = null;
-        }
-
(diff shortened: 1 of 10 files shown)


blynch11p:
This doesn't handle the case where a user already has a URL from a previously shared tag that does not have the entity id in the URL.

The link still loads the results for the tags, but the subscribe button shows and does not work.

mleeBoeing:
Removed the subscribe/unsubscribe link if the tag id isn't present. URL from a previously shared tag will not show the subscribe/unsubscribe

blynch11p:
Subscribe/Unsubscribe link works.
Link does not show when the URL does not contain the ID for backwards compatibility with previously shared tag links.
:+1:

lsemesky:
when going direct to link for a subscribed tag (eg) http://localhost:8000/dist/#/home/////tag_0/3 the text originally says (subscribe) and then switches to (unsubscribe) once data loads. Perhaps it's better not to show anything until the subscription data loads? I don't have a strong feeling one way or another though so approved if Brian doesn't think it matters :)
:+1:

blynch11p:
Good point, @lsemesky.

@mleeBoeing, if you can stop the text from showing prior to the data loading, with reasonably little effort, I would prefer that.  

The good news is that it doesn't appear you can double subscribe.

mleeBoeing:
Do not merge. Found a bug while submitting a new listing.

mleeBoeing:
For the subscription data being hidden until the data loads, it's going to take some effort. Will need to make the ajax call in the webapi to be synchronous. However, the async property is being deprecated on Firefox and Chrome.

blynch11p:
I'm leaning towards you being correct that this is a bit of effort.  I haven't found an element (within props or state) that can be looked at for subscriptions having finished loading.  Looking at if `element.stop` exists works when there is at least one subscription  (element.stop becomes undefined), but if there are no subscriptions it never becomes undefined.

mleeBoeing:
Had a bug where create/edit listing page would show [object, object] for the Tag input. Therefore, not allowing you to save or submit. Pushed fix in. Requesting review again.

app/js/components/discovery/DetailedQuery.jsx
```diff
@@ -192,18 +192,19 @@ var DetailedQuery = React.createClass({
...
-        if (this.props.data.tagId.length > 0) {
-            var subscribeLink = <a className="tag_subscribe" onClick={(e) => {this.subscribeToTag(e)}} >Subscribe</a>;
-        } else {
-            subscribeLink = null;
-        }
-
-        if (this.state.subscriptionStore && this.state.subscriptionStore.length > 0) {
+      var subscribeLink = null;
+      //last conditional is a fix for subscription store not being null even if it hasn't loaded yet
+        if (this.state.subscriptionStore && this.state.subscriptionStore.length > 0 && this.state.subscriptionStore[0].entity_description) {
```
lsemesky: maybe entity_description should be changed to entity_id or something we check for later so that we aren't adding an additional api dependency

lsemesky:
see any issues with the latest @blynch11p ?

blynch11p:
@lsemesky, nope!!!  Good stuff.

PR #810, opened 2017-06-13
Fix bug where adding new category messes up with subscription links

mleeBoeing:
Subscription links currently break when a new category gets added. This is a fix for that.


blynch11p:
:+1:

mannyrivera2010:
:+1:

PR #816, opened 2017-06-27
Launch notice of requirements

blynch11p:
Opens a notice after clicking Launch providing guidance to the user if there are requirements in place for using the application.  A 30 second timeout is implemented to reduce the number of clicks needed by a user and should provide an acceptable amount of time for the user to see a page has not loaded, return to AML, and read the notice.

screenshot from 2017-06-27 16-59-51

The dialog box will NOT open if the Requirements field for an Application is either empty or None.


blynch11p:
This work is being done in https://github.com/aml-development/ozp-center/pull/846

PR #820, opened 2017-07-10
Sub long cat names

blynch11p:
… subscribe link

Edit a category name or two to make it much longer (Center Settings -> Categories):

Make sure the category names and (un)subscribe links on Center don't interfere with each other.

Also, resize the browser window (max) and shrink it width wise and verify nothing looks 'off'

@mleeBoeing @J-Fid @lsemesky @rkenyon1969


mleeBoeing:
:+1:

PR #824, opened 2017-07-14
Sort by

mleeBoeing:
Add interactive sort button to enable users to change the sort order of listing results: Most Popular, Category, Tag, Search.

## **You will need this backend: https://github.com/aml-development/ozp-backend/pull/325**

You will need to do a 'make reindex_es'

### DIFF
--- a/app/js/components/discovery/index.jsx
+++ b/app/js/components/discovery/index.jsx
@@ -35,8 +35,13 @@ var timeout;
 
 
 var FILTERS = ['categories', 'type', 'agency', 'tags'];
-var sortOptions = ['Newest', 'Title: A to Z', 'Title: Z to A', 'Rating: Low to High', 'Rating: High to Low'];
-var searchSortOptions = ['-approved_date', 'title', '-title', ['avg_rate', '-total_votes'], ['-avg_rate', '-total_votes']];
+var sortOptions = [
+    {option: 'Newest', searchParam: '-approved_date'},
+    {option: 'Title: A to Z', searchParam: 'title'},
+    {option: 'Title: Z to A', searchParam: '-title'},
+    {option: 'Rating: Low to High', searchParam: ['avg_rate', '-total_votes']},
+    {option: 'Rating: High to Low', searchParam: ['-avg_rate', '-total_votes']}
+];
 
 var areFiltersApplied = (state) => {
     return _.reduce(FILTERS, function (memo, filter) {
@@ -79,6 +84,7 @@ var Discovery = React.createClass({
         this._searching = true;
         this.setState({
             queryString: evt.target.value,
+            orderingText: 'Sort By',
             currentOffset: 0
         });
     },
@@ -112,27 +118,13 @@ var Discovery = React.createClass({
     },
 
     onSortChange(order) {
+        var me = this;
         if (order != this.state.ordering) {
-            switch (order) {
-                case searchSortOptions[0]:
-                    this.setState({orderingText: sortOptions[0]});
-                    break;
-                case searchSortOptions[1]:
-                    this.setState({orderingText: sortOptions[1]});
-                    break;
-                case searchSortOptions[2]:
-                    this.setState({orderingText: sortOptions[2]});
-                    break;
-                case searchSortOptions[3]:
-                    this.setState({orderingText: sortOptions[3]});
-                    break;
-                case searchSortOptions[4]:
-                    this.setState({orderingText: sortOptions[4]});
-                    break;
-                default:
-            }
-
-            this.setState({ordering: order, currentOffset: 0});
+            sortOptions.forEach(function(element) {
+                if (order == element.option) {
+                    me.setState({orderingText: element.option, ordering: element.searchParam, currentOffset: 0});
+                }
+            });
         }
     },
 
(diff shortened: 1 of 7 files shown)


app/js/components/discovery/index.jsx
```diff
@@ -485,8 +553,16 @@ var Discovery = React.createClass({
         }
 
         var searchLink = `${CENTER_URL}/#/home/${encodeURIComponent(this.state.queryString)}/${(this.state.categories.length) ? encodeURIComponent(this.state.categories.toString()).replace(/%2C/g,'+') : ''}/${(this.state.type.length) ? encodeURIComponent(this.state.type.toString()).replace(/%2C/g,'+') : ''}/${(this.state.agency.length) ? encodeURIComponent(this.state.agency.toString()).replace(/%2C/g,'+') : ''}/${(this.state.tags.length) ? encodeURIComponent(this.state.tags.toString()).replace(/%2C/g,'+') : ''}/${(this.state.tagId.length) ? encodeURIComponent(this.state.tagId.toString()).replace(/%2C/g,'+') : ''}`;
+
         return (
             <section className="Discovery__SearchResults">
+                <SelectBox className="SelectBox sortBy" label={this.state.orderingText} onChange={this.onSortChange}>
```
lsemesky: As Brian stated, you can use a map here (see app/js/components/discovery/Types.jsx for an example

app/js/components/discovery/index.jsx
```diff
@@ -34,6 +35,8 @@ var timeout;
 
 
 var FILTERS = ['categories', 'type', 'agency', 'tags'];
+var sortOptions = ['Newest', 'Title: A to Z', 'Title: Z to A', 'Rating: Low to High', 'Rating: High to Low'];
```
lsemesky: I am not thrilled that these 2 arrays are separate because you are relying on them both to stay the same order independently. Perhaps one array with the following structure would be better?
```
[
    {
        name: 'Newest',
        value: '-approved_date'
    },
    {
        name: 'Title: A to Z',
        value: 'title'
    },
    ...
]
```

app/js/components/discovery/index.jsx
```diff
@@ -106,6 +111,31 @@ var Discovery = React.createClass({
         this.setState({ agency, currentOffset: 0 });
     },
 
+    onSortChange(order) {
+        if (order != this.state.ordering) {
+            switch (order) {
+                case searchSortOptions[0]:
```
lsemesky: since these are already in an array you are probably better off doing this programmatically by looping through the array.

app/js/components/discovery/index.jsx
```diff
@@ -440,6 +477,30 @@ var Discovery = React.createClass({
           }, 500);
         }
     },
+
+    sortMostPopular(order) {
+        DiscoveryPageStore.sortMostPopular(order);
+
+        switch (order) {
```
lsemesky: same deal as above.. programmatically is better

app/js/components/discovery/index.jsx
```diff
@@ -450,6 +511,13 @@ var Discovery = React.createClass({
         return (
             <section className="Discovery__MostPopular" key="Discovery__MostPopular">
                 <h4>Most Popular</h4>
+                    <SelectBox className="SelectBox sortBy" label={this.state.orderingText} onChange={this.sortMostPopular}>
+                        <option className="sortBy" value="newest">{sortOptions[0]}</option>
```
lsemesky: As Brian stated, you can use a map here (see app/js/components/discovery/Types.jsx for an example.

app/js/components/discovery/index.jsx
lsemesky: I like how you did the value programmatically here instead of hard-coding as the above.
Also since you print the select twice (it is the same, right?) you can refactor out to a function.

app/js/stores/DiscoveryPageStore.js
```diff
@@ -61,6 +61,64 @@ var DiscoveryPageStore = Reflux.createStore({
...
+                    return (a.avgRate < b.avgRate) ? -1 : 1;
+                }
+            }
+        });
+
+        return arr;
+    },
+
+    sortMostPopular(order) {
+        switch (order) {
```
lsemesky: so what you could do here to make this programmatically, is if you change the structure to a single array of objects, you could include a sortMethod parameter and assign the appropriate function to each value, eg (warning: pseudocode, may not work as written)
```[
    {
        name: 'Newest',
        value: '-approved_date',
        sortMethod: function(array) { return this.sortNewest(array);}
    },
    {
        name: 'Title: A to Z',
        value: 'title',
        sortMethod: function(array){ return this.sortAlphabetically(array, 'asc');} //technically you don't have to provide the second parameter here... javascript will still call the function even if the number of params in the signature is different
    },
    ...
]
```

then you could loop through the array and just call the sortMethod once the appropriate array entry is found

app/js/stores/DiscoveryPageStore.js
```diff
@@ -61,6 +61,64 @@ var DiscoveryPageStore = Reflux.createStore({
...
+                _mostPopular = this.sortAlphabetically(_mostPopular, "asc");
+                break;
+            case "titleZA":
+                _mostPopular = this.sortAlphabetically(_mostPopular, "desc");
+                break;
+            case "newest":
+                _mostPopular = this.sortNewest(_mostPopular);
+                break;
+            case "ratingLoHi":
+                _mostPopular = this.sortRating(_mostPopular, "asc");
```
lsemesky: just a note about javascript:
 technically you don't have to provide the second parameter here since including it for ascending sort doesn't do anything in the function... javascript will still call the function even if the number of params in the signature is different, it just goes by the first function in scope matching the name.

rkenyon1969:
- Suggest increasing .sortBy  width from 140px to 190px so that all options are readable
- also, should we have a discussion on default option based on results?

blynch11p:
Sort By functions as expected on both the main Center screen and Search result screen. . :+1:

rkenyon1969:
**Issue:** sort by state does not change or reflect accurately sort order of returned results when clicking on a category from center home:


**STR:**
1. On "Center Home" sort "Most Popular" by "Ratings: Low to High"
2. Click on "Communication" Category in left navigation

**Results:** return results are in default state but sort by state is still "Ratings: Low to High" which does not match order of returned results

**Expected Results:** sort by state matches order of return results or state defaults to "Sort By"

blynch11p:
@rkenyon1969, good catch.  @mleeBoeing, the same happens with filtering on Listing Type or Organizations.  The Search box _does_ reset the state of the Sort By menu.

PR #826, opened 2017-07-20
Update url after clicking a Similar Listing

blynch11p:
There was also a side-effect of the URL not updating that would make a Similar Listing that was loaded be stuck on the tab (Details/etc) the Quick View was on before loading that Similar Listing.

To test:
Use user bettafish
Prior to pulling branch, open a listing from Center, refresh page, click Details tab, after Similar Listings loads, click a listing to load:
Notice the URL does not change
In some cases, when clicking the tabs in quick view after loading the Similar Listing the tabs will not function as expected (the url will update, but not the display)

Pull the branch, following above steps above while making sure the URL updates as expected after choosing a Similar Listing, and the listing goes to the Overview tab.
Also verify that the tabs in the Similar Listing behave as expected.

@mleeBoeing @lsemesky @rkenyon1969

### DIFF
--- a/app/js/components/discovery/RecommendedListingTile.jsx
+++ b/app/js/components/discovery/RecommendedListingTile.jsx
@@ -30,9 +30,10 @@ var RecommendedListingTile = React.createClass({
         }
     },
 
-    loadRecommendation: function (recommendation) {
+    loadRecommendation: function (recommendation, href) {
         OzpAnalytics.trackRecommender(listingMessages['recommender.similar'], recommendation.title);
         CurrentListingStore.loadListing(recommendation.id);
+        //window.history.pushState('','', href);
     },
 
     render: function () {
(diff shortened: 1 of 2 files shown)


mleeBoeing:
:+1:

Tested with bettafish. Able to see the URL change when clicking on a new listing under Similar Listings. Does not happen when I pull a different branch.

rkenyon1969:
1. Pulled branch
2. reviewed code
3. tested

- Use user bettafish :+1: 
- Prior to pulling branch, open a listing from Center, refresh page, click Details tab, after Similar Listings loads, click a listing to load: Notice the URL does not change :+1: 

- Pull the branch :+1: 
- following above steps above :+1: 
- making sure the URL updates :+1: 
- and the listing goes to the Overview tab. :+1: 
- Also verify that the tabs in the Similar Listing behave as expected. :+1: 

:+1:

PR #827, opened 2017-07-20
Update toggles in Quickview->Administration visually

mleeBoeing:
The toggles for Preferences now display ON/OFF.  The toggles in the Administration tab for an App should follow the same design.

### DIFF
--- a/app/js/components/createEdit/form/Toggle.jsx
+++ b/app/js/components/createEdit/form/Toggle.jsx
@@ -19,16 +19,13 @@ var Toggle = React.createClass({
           <div id={this.props.id}>
               <label className="switchLabel">{this.props.label}</label><br />
               <p className="small">{this.props.description}</p>
-              <label className="createEditSwitch">
-                  <input ref="checkbox"
-                         type="checkbox"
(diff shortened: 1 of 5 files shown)


blynch11p:
Those buttons are huge.  Recommend making them a little closer to the size they originally were.
screenshot from 2017-07-20 17-50-11
screenshot from 2017-07-20 17-48-32

app/js/components/createEdit/form/Toggle.jsx
```diff
@@ -14,21 +14,19 @@ var Toggle = React.createClass({
       if (!this.state.value) {
           fullExplanation = isEnabled + this.props.explanation[0];
       }
+    //   console.log(this.props.id1);
```
blynch11p: Unneeded

blynch11p:
:-1: 

The 'singleton' toggle on Edit Listing doesn't not actually save state.  Private Listing appears to however.
There's also one commented out console.log in the code.

mleeBoeing:
Fixed

blynch11p:
Corrections fixed the issues with Singleton and side-effects I was seeing with the Admin tab. :+1:

PR #828, opened 2017-07-20
Stop and show warning if a category has existing listings and is abou…

mleeBoeing:
Acceptance Criteria -
If category that is being deleted is the only category listed to some of the app listings the admin should be made aware in order to reassign a new category to it.

### DIFF
--- a/app/js/mixins/AjaxMixin.js
+++ b/app/js/mixins/AjaxMixin.js
@@ -18,7 +18,7 @@ var AjaxMixin = {
             errors: response ? response.errors : null,
             errorStatus: status,
             errorMessage: response ?
-                    (response.message || response.responseText) :
+                    (response.message || response.responseText || response.detail) :
                     response.responseText
         });
     }


blynch11p:
Is it possible to get the "cannot delete" notice prior to clicking the modal's Delete? (is it easy to correct the "Can not" to be "Cannot"?)
Also, after the notice pops up the delete button is still there.

However, this does work and stop the user from removing a category.  And I know this won't be used by a lot of people.

@rkenyon1969, thoughts on if this is fine as-is if the level of effort involved regarding my remarks above is more than a few minutes of work?

mleeBoeing:
Error message change PR link: https://github.com/aml-development/ozp-backend/pull/330

rkenyon1969:
I agree with Brian's points. I suggest doing the following:

1. User selects category and then delete
2. if(listings) then modal with just warning message (w/ "Cannot") and cancel button (not delete button)
3. If(!listings) then normal delete modal

Alternative Message: "The category xxx cannot be deleted. The category has xx listings associated with it."

Thoughts?

mleeBoeing:
If this feature will be used only once or twice, in my opinion the cost benefit is not worth.

blynch11p:
:+1:

rkenyon1969:
I agree with Brian's last point. :+1:

PR #829, opened 2017-07-21
Sort alphabetically prior to sorting ratings to keep dataset the same…

mleeBoeing:
Mainly for sorting most popular -> Ratings: High to Low. This is to keep the listing store sorted based on front end in the beginning. Also, to keep the dataset in the same order each time prior to sorting by ratings.


blynch11p:
Good idea for putting the array in a 'default' sort prior to sorting by rating so the sort outcome is always the same. :+1:

PR #830, opened 2017-07-25
Remove delete confirmation message if error occurs

mleeBoeing:
If you ARE NOT allowed to delete the error message, then remove the "Are you sure that you would like to delete the category "Tools"? ** and the Delete and Cancel button.  If you ARE allowed to delete the category , then remove the message "Cannot delete category Tools" part of the error message AND make sure the delete functionality actually works.


blynch11p:
Better :+1:

PR #831, opened 2017-07-26
Notification resend

blynch11p:
…st notification.

To test:
- Go to Center Settings -> Notifications
- In the Past Notifications column, there will be a Message icon next to the Create/Expired dates.
Click the message icon.
- The message will populate the Notification Text textearea in the Send a Notification column.
- Complete the notification form (select a date) and click Send.

Verify the Notification is sent with the correct text.

Other tests:
Make sure a pre-populated notification can be edited and sent correctly.
Verify the Reset button functions correctly.
Verify form resets after sending a notification (both created from pre-populate and from manually typing)

@mleeBoeing @lsemesky @rkenyon1969

### DIFF
--- a/app/js/components/management/mall/notifications/CreateNotification.jsx
+++ b/app/js/components/management/mall/notifications/CreateNotification.jsx
@@ -31,10 +31,14 @@ var CreateNotification = React.createClass({
         return {
             uuid: uuid(),
             date: null,
-            message: '',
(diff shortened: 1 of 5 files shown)


mleeBoeing:
:+1:
Was able to verify the past notification would copy and send correctly. Form resets as expected afterwards. 
Reset button also works as expected.

PR #832, opened 2017-07-26
Fix Lo to Hi issue with total ratings

mleeBoeing:
Fix sort by issue when doing Ratings -> Low to High. If the ratings were equals, the tie breaker would be the total # reviews (most would show up first). Now it should be the opposite (most total # reviews should show last).

### DIFF
--- a/app/js/components/discovery/index.jsx
+++ b/app/js/components/discovery/index.jsx
@@ -298,6 +298,9 @@ var Discovery = React.createClass({
 
 
     onStoreChange() {
+        if (!this.state.searching) {
+            this.setState({orderingText: 'Sort By'});
+        }
         this.setState(this.getInitialState());
     },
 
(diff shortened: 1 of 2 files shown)


blynch11p:
Low to High sort order appears to be more in line with what's expected.
:+1:

rkenyon1969:
1. Pulled branch
2. reviewed code
3. tested

- most total # reviews should show last 
- L-H is the reverse order of H-L/MP :+1: 

:+1:

rkenyon1969:
@J-Fid - what do you mean by "when a change is made"? - thanks,

J-Fid:
@rkenyon1969 Reviewing a listing causing its rating to change.  Michael just put in a fix for the the drop down bar, and I'll discuss it with him when he comes back from lunch.

PR #834, opened 2017-08-02
Separating the frontend storefront calls to use the new endpoints

lsemesky:
to run locally:
make sure you checkout the master backend branch
make sure you pull the latest into the master backend branch
make dev on backend if needed
make reindex_es on backend if needed

checkout this frontend branch

to test:
open developer tools, the network tab
Make sure that when storefront loads there are no references to 
http://127.0.0.1:8001/api/storefront/

and instead are references to the following endpoints
http://127.0.0.1:8001/api/storefront/featured
http://127.0.0.1:8001/api/storefront/recent
http://127.0.0.1:8001/api/storefront/most_popular 
http://127.0.0.1:8001/api/storefront/recommended

Ensure that recommended recomendations still only appear for the beta users (eg. bettafish)

Ensure that store front, search, and clicking on apps still function

### DIFF
--- a/app/js/actions/ListingActions.js
+++ b/app/js/actions/ListingActions.js
@@ -10,7 +10,11 @@ var ListingActions = createActions({
     fetchCounts: null,
     fetchAllChangeLogs: null,
     fetchAllChangeLogsByID: null,
-    fetchStorefrontListings: null,
+    fetchStorefrontListings: null, //depricated
+    fetchFeaturedListings: null, 
+    fetchRecentListings: null,
+    fetchMostPopularListings: null,
+    fetchRecommendedListings: null,
 
     fetchById: null,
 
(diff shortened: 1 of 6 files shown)


blynch11p:
Two things I'm noticing:

The important one 
- I'm not getting the images for Recommended For You.  The image is available in the API call, but it's not in the  tag

The weird thing I see (that doesn't seem to affect performance however):
- /api/metadata takes less than 1 second to load before pulling this branch
- It's taking 3 seconds after pulling the branch.  

Otherwise, I <3 that visually this is definitely faster. Pre-fetch times were 5-10s, post-fetch are 4 or less before something is on the screen.

mannyrivera2010:
image

mannyrivera2010:
:+1:    It looks likes it works.

blynch11p:
The /metadata call is still confusing, but it could be tied to data throughput since there are now ~10 calls made simultaneous vs 7.

Code looks good.
Calls are broken up into small result sets.
Search functions as expected.
All images load.

:+1:

PR #844, opened 2017-08-21
Fix bug where apps not reflecting the correct rating in similar listings

mleeBoeing:
Similar listings would not show the correct ratings in the quick view modal. The top left would update however the bottom bar wouldn't update. 

Sign in as bettafish
Go to Air Mail
Go to Air Mail 2 and add a review
Go back to Air Mail and you should see the updated review


blynch11p:
Code looks good
Functions as expected

:+1:

rkenyon1969:
1. pulled branch
2. reviewed code
3. tested

:+1:

PR #846, opened 2017-08-23
Notify user when they're leaving OZP

mleeBoeing:
- Enter a delay between clicking Launch and opening the new browser tab/window.  Enough time (~3sec) for the user to see a modal has been displayed that informs the user they are leaving OZP.
- Modal should have a generic message about leaving OZP and list the requirements needed.
- Move Usage Requirements (currently in Details Tab) to the Overview tab above the description.

When launching an app and it opens a new browser tab/window, the browser will think it's a popup and will block it. Popup blocker will possibly need to be disabled.


mleeBoeing:
Working on more functions, PR is not ready. Removing reviewers.

blynch11p:
Actually 'using' this new feature, I have some questions/comments/recommendations.

- 3 seconds is fine when all goes well and the browser correctly opens the window
 -- Is there a way to detect popup blocked?  So, if after 3 seconds, the pop up was blocked, don't close the dialog box, and put a note "click here to open".
-- If I want to "Not show this dialog box again", once I interact with the dialog box, I think the timeout should be cancelled, so you have ample time to check the box, and click OK.  (unless clicking the check box automatically stores the information).

I'm starting to wonder if we want the 3 seconds delay for opening the new URL, but ~10 seconds for closing the dialog.  The timing seems to be rushed.

mleeBoeing:
Merged the requirements branch into this branch. Also have added the popup detector. If application launch is blocked due to the browser's popup blocker, an alert will show which will give you the option to manually launch the application. I have also increased the warning modal to 30 seconds from 3 seconds. Automatic launching of the application is staying the same at 3 seconds.

app/js/components/CenterLaunchLink.jsx
```diff
@@ -39,17 +47,81 @@ var CenterLaunchLink = React.createClass({
...
+    },
+
     render: function() {
-        var listingLaunchFn = ListingActions.launch.bind(null, this.props.listing),
-            { className, ...otherProps } = this.props,
+        var { className, ...otherProps } = this.props,
             linkClassName = className ? className + ' btn' : 'btn';
+        var launchWarning = this.state.launchWarning;
 
+        var launchModal = this.state.currentUser && this.state.currentUser.leavingOzpWarningFlag ? (<Modal ref="modal" className="LaunchConfirmation" size="small" title="Launch Requirements Notice" onCancel={this.modalConfirmation}>
```
blynch11p: `title` should now be `modaltitle`

app/js/services/ListingService.js
```diff
@@ -244,9 +246,37 @@ ListingActions.saveReviewResponse.listen(function (listing, review) {
...
+    if (timeout) {
+        var application;
+
+        setTimeout(function() {
+            application = window.open(listing.launchUrl);
+
+            if (application == null || typeof(application) == 'undefined') {
+                swal({
+                    title: "Pop-up blocked",
+                    text: "Click launch to open application manually.",
```
blynch11p: Needs a text change.

The button is not labelled "launch".
Maybe "Click to launch application manually"
or
"Click 'Launch Application' to open manually"
or fix the button text.

app/js/components/CenterLaunchLink.jsx
mleeBoeing: Updated

app/js/services/ListingService.js
mleeBoeing: Updated

blynch11p:
Dialog box pops up after clicking launch. ✔️ 
Delay is present before opening URL ✔️ 
The user can open the URL if the browser blocks the popup ✔️ 
The dialog closes automatically ✔️ 
The user can opt in/out of the dialog ✔️ 

👍

rkenyon1969:
1. pulled branch(es)
2. reviewed code
3. tested

**Question:** @mleeBoeing - is there a negative to reducing timeout from 30000 to 800?

:+1:

Issue #847, opened 2017-09-05
Tab Flow Indicator


mleeBoeing:
Modify "No Thanks" button to depict when the user has tabbed to. Similar to "Start the Tour" button turns blue once focused on. See attached images.

![image](https://user-images.githubusercontent.com/26389772/30084125-22fa36fe-925f-11e7-800d-5079bab7345f.png)![image](https://user-images.githubusercontent.com/26389772/30084118-1c0e87f0-925f-11e7-8989-5c52e5eaec10.png)

flesher1:
Added class to commons.scss so that the "No thanks" and "End Tour" buttons turned blue like the "Start the tour" button.

PR #851, opened 2017-09-07
Make tour button blue

flesher1:
In center tour, hover mouse over 'No thanks' button. Button should turn blue. If tabbed to the button should turn blue. Similar to the 'Start the tour' button. Following ticket #847 

Pull https://github.com/aml-development/ozp-react-commons/pull/93 first


app/styles/main.scss
```diff
@@ -45,6 +45,14 @@ html, body {
   padding-left:0;
 }
 
+// class used to change color of buton to blue when hovered or focus
```
blynch11p: Take a look at
https://github.com/aml-development/ozp-react-commons/blob/master/app/styles/commons.scss#L97

PR #872, opened 2017-10-02
Disabled listing fix

flesher1:
Issue https://github.com/aml-development/ozp-center/issues/868 

The wording is what I made, please review. The button color and text is also not final.

There is a different warning for an admin and for a regular user. A user will get an error and a admin will get a warning and the listing will remain open. Please open a disables listing as both a regular user and admin to test fully.


app/js/components/quickview/index.jsx
```diff
@@ -82,6 +83,19 @@ var Quickview = React.createClass({
         var { shown, listing } = this.state;
         var ActiveRouteHandler = this.getActiveRouteHandler();
         var owners, tabs;
+
+        if (listing){
+            if ((!currentUser.isAdmin()) && (listing.isEnabled == false)){
+                //if (listing.isEnabled == false){
```
blynch11p: This line and line 93 can be removed.

blynch11p:
Tested:
Regular users see a notice that the listing is no longer available
'Admin' users are notified that the listing is disabled for regular users, but the admin can still view the listing

But...
Any user (org steward/listing owner) that can enable/disable a listing, should also see the notice that an Admin sees.  Meaning, any user type for a Listing that can see the Admin tab, should not be blocked from seeing the listing.  [this was discussed/confirmed with @rkenyon1969]

blynch11p:
I like it.

Deleted, for all users, does not show listing, displays dialog saying listing deleted

Disabled:
Regular users see a notice that the listing is no longer available
Admin/Org Steward/Owner users are notified that the listing is disabled for regular users, but the admin can still view the listing

👍

rkenyon1969:
1. pulled branch
2. reviewed code
3. tested

:+1:

PR #873, opened 2017-10-04
Stop deleting reviews

flesher1:
https://github.com/aml-development/ozp-center/issues/869

This removes the Delete button from reviews in listings that the user is not a steward of or if the user is not the author.


blynch11p:
Edit icon only displays for Reviews created by the user or for Org Stewards of the listing
👍

mleeBoeing:
:+1:
Not able to see the edit icon as wsmith

image

PR #876, opened 2017-10-11
Disabled listing fix

flesher1:
Checked with users Johnson, Obrien, and Bigbrother. User can submit/delete listing and go through entire process without receiving popup (excluding when deleting as expected). Popup only appears on the first render of the listing. a popup will appear for an admin with one button when the listing is deleted, or pending deletion. The popup will stay open for the pending deletion and close automatically for deleted. An Admin will receive a warning with two buttons (One enable and one close warning) when a disabled listing is opened. the enable button will enable the listing. both buttons will keep the listing open. All admin functions apply to listing owners.

For a regular user, the user will receive an error message if the listing is deleted, disabled or pending deletion. The listing will close.


blynch11p:
Appropriate notices are displayed for regular/admin users.
Notice doesn't continuously show when switching tabs while viewing a disabled listing as an admin.
Notice displays when switching to a new listing.

👍

mleeBoeing:
Need to squash the commits

flesher1:
@mleeBoeing I will squash when I merge. I don't plan to merge all the commits to master.

rkenyon1969:
1. pulled branch
2. reviewed code
3. tested

:+1:

PR #878, opened 2017-10-18
JIRA-750: Fixing the style so the private icon does not overlap

blynch11p:
To test:

Prior to pulling branch, log in to Center using obrien.
Find a Listing under the Minipax org (ex: Wolf Finder)
Open the listing and click the edit button.
Scroll down to the "Private Listing" section and enable the listing as private.
Click Save.
Return to Center, find the Minipax listing that was edited, open the listing, and see that the Lock icon is overlapping the Minipax text.

Pull this branch.
Restart center server
Refresh the browser.
The icon should no longer be overlapping the text.


mleeBoeing:
Prior:
image

After:
image

:+1:

rkenyon1969:
:+1:

PR #890, opened 2017-10-27
fix(listing_loading): update loading a listing to prevent tabs from b…

blynch11p:
…eing visible before they can be clicked

JIRA-789

The issue was that before a listing loaded, the user could click the Administration tab, which would close the listing.

This was a side effect of trying to show the listing before it was loaded.

There's an additional check in place that the listing is available (listing.title).

The styling of the loading has also been update to include the new loading icon.

To test:
Load a listing from Center, and make sure the dialog shows 'loading' and not tabs before the listing is loaded.
Refresh the page to verify the loading message displays.
Also, create a listing from Create/Edit, Save, and view the preview to verify the preview works.


mleeBoeing:
:+1:
Can see the loading icon while it loads

rkenyon1969:
:+1:

PR #780, opened 2017-03-30
Made 'Categories Section' to be more descriptive for 508 users

mleeBoeing:
Made 'Categories Section' to be more descriptive for 508 users

PR #891, opened 2017-10-27
fix(Management): correct credential checking in listing management

flesher1:
Add system store to app startup to account for bookmarked pages
Add processing in Mixin to translate Title to shortName

Closes #780 (DI2E)

Requires https://github.com/aml-development/ozp-react-commons/tree/OrgSteward_check_correction


blynch11p:
Appears to be working as expected

👍

PR #894, opened 2017-11-01
fix(sort): fix # reviews when sorting

mleeBoeing:


blynch11p:
Sorting works as expected now
👍

flesher1:
Most Popular is now sorted properly.
:+1:

PR #897, opened 2017-11-09
Refactor listing status icon to its own component

flesher1:
AMLNG-215 and AMLNG-820

There is a new file created called ApprovalStatusIcons which is the source of all icons relating to approval status. It serves as the single point of reference and a location for changes to be made to icons if necessary. 

Additionally, in there is now a new set of filters on the Listing Management/My listings page. There are two new Pending filters which allow the user to see listings that are waiting for Organization approval and AML admin approval. There is also a Deleted filter. The filters on the My Listings tab now work in the same way that the All Listings and Org Listings tabs work. 

To test, check functionality of these modifications against the existing master. The notable areas to check are listing management page (All tabs and filters) and administration tabs on listings. Also, the My Listings tab in listing management has went through interface changes. Please review it for new functionality described above.


blynch11p:
Take a look at making the icon for Pending AML Approval  (also seen as Pending, Center in the Admin tab of a listing) consistent regardless of user type.  
The icon in the left menu using the 'loading' icon for both pending options, but an admin will see an [!] for Listing in the Pending AML Approval state.  Visually it's confusing when All Listing are shown, and the user sees an [!] but the "Needs Action" menu item has 0 items.


screenshot from 2017-11-09 16-34-00

app/js/components/listing/ListingTile.jsx
```diff
@@ -153,11 +146,17 @@ var EditedDate = React.createClass({
 
 var InfoBar = React.createClass({
     render: function () {
-        var listing = this.props.listing;
+        var listing = this.props.listing,
+        user = this.props.user;
+        var lockStyle = {
```
mleeBoeing: Shouldn't this be in a css file?

flesher1:
Moved styling and made change to Needs Action Icon as requested. Below is a picture of the change. Know that this is the expected outcome, and the NavBar(below the listing) is the same on each screen and therefore cannot be modified within the scope of this ticket.

image

blynch11p:
Everything looks good.
Explanation of the status icons makes sense.
👍

rkenyon1969:
@flesher1 - we need to revisit the workflow states and determine as a team what is the workflow and how should that be visually presented to the end-user. This is out of the scope of this task. I will create a Jira ticket to address.

:+1:

PR #907, opened 2017-11-22
fix(launch popup): properly close modal when clicking OK

irizvifearless:
add close function on CenterLaunchLink

closes AMLNG-833

**Description**
Scrollbar disappears once you press the OK button on the launch notice modal. Issue most likely comes from the fact that the OK button is calling a method and forgot to call the closeModal function.

**Acceptance Criteria**
Go to Homepage
From Thumbnail view, Launch App from thumbnail view
Click OK to close modal window
EXPECTED - App returns to normal state with functional scrollbar
RESULTS - Scrollbar no longer allows scroll

**How to test code**
Pull amlng833 from ozp-center

### DIFF
--- a/app/js/components/CenterLaunchLink.jsx
+++ b/app/js/components/CenterLaunchLink.jsx
@@ -79,12 +79,17 @@ var CenterLaunchLink = React.createClass({
 
     modalConfirmation: function(){
         var profile = this.state.currentUser;
+        var { listing } = this.getQuery();
         profile.leavingOzpWarningFlag = this.state.leavingOzpWarningFlag;
 
         ProfileActions.updateProfileFlags(profile);
         ProfileActions.fetchProfile(profile.profileId);
 
         this.reset();
+
+        if(!listing) {
+            this.close();
+        }
     },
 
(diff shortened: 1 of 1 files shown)


mleeBoeing:
:+1:
Fixes the issue

blynch11p:
There's actually a side-effect of this*, that doesn't actually cause any major issues, and does fix the issue that causes the user experience to suffer, so 👍 

* When opening the Quickview modal, the scrollbar goes away. When clicking launch, the scrollbar remains away when the launch notice displays.  When closing the launch notice, but not the Quickview modal, the scrollbar comes back, so now the user can scroll in a place they previously could not.  BUT, this doesn't affect user experience, so it's not a concern.

irizvifearless:
@blynch11p good catch, I see that issue now.  I committed a potential fix for that -- do not call the close function if the launch popup came from a quickview.  I check for the listing ID in the URL query string to determine if we are in quickview or not, the same way the render function does.  Let me know if this is acceptable.

blynch11p:
Cool.  Yup, that works.

👍

PR #912, opened 2017-11-30
fix(listing management): fix the listings not updating within the gri…

mleeBoeing:
Fixes this issue: After you edit/delete a listing, the Listing Management table view doesn't seem to be updating at all until a hard refresh occurs. The tabs on the left side also need a hard refresh to update.


blynch11p:
Numbers in the left side bar update after deleting listings from grid view 👍

mannyrivera2010:
:+1:

PR #913, opened 2017-12-05
refactor(carousel): Remove and Replace caroufredsel

flesher1:
Replace the current Carousel:
  This removes the current Carousel from center and creates a custom carousel.
  The purpose of this was to closely recreate the functionality of the current carousel.
  The carousel is React based instead of JQuery. It utilizes JQuery for animations.
  No new features have been added. Some minor functionality differences exist.

AMLNG-842

**Description**
The current carousel in use (caroufredsel) has limited some of what we've tried to do lately (recommender feedback and removing listings on the fly), so we're going to take some time to create our own.

**Acceptance Criteria**
* Similar behaviour to current carousel to maintain current user experience (click arrows to move slider/carousel, animated)
* Works with react style prop/state updates

**How to test code**
Pull this branch. You will need to wipe out your node modules and run npm install to get rid of Caroufredsel. After starting center, there are carousels on the main homepage and within the listing containing screenshots. Confirm that they function the same as the current carousel (caroufredsel).


app/js/components/carousel/index.jsx
```diff
@@ -2,74 +2,141 @@
...
 
-    componentWillUnmount: function () {
-        if (this._$carousel) {
-            this._$carousel.trigger('destroy');
+    render: function() {
+        var items;
+
+        //is this bad practice in react?
+        if (this._currentElement.ref==="featured"){
+            items = this.state.slides.map((slide, index) =>
```
blynch11p: Since it appears the only difference here is the class, just try something like

```
            items = this.state.slides.map((slide, index) =>
                var classes = "carousel__slide" + (this._currentElement.ref==="featured" ? " carousel__slide--featured" : "");
                return(
                    <div className={classes}>
                        {slide}
                    </div>
               );
           );
```

(the correctness of syntax is questionable)

Also, another option to figure out what list you're in (and this isn't necessarily better than _currentElement.ref) is
`this.state.slides[0].type.displayName`
and the result would be `FeaturedListingTile` (or one of the others)

app/js/components/carousel/index.jsx
```diff
@@ -2,76 +2,132 @@
...
+                className="carousel__arrow carousel__arrow--right"
+                onClick={this.props.onClick}
+                >
+                    <i className="center__icon icon-caret-right-white"></i>
+            </a>
+        );
+    }
+});
 
+var CarouselSlide = React.createClass({
```
blynch11p: No longer needed with the latest change

app/js/components/carousel/index.jsx
```diff
@@ -2,76 +2,132 @@
...
+    render() {
+        return(
+            <div className="carousel__slide">
+                {this.props.slide}
+            </div>
+        );
+    }
+});
+
+var FeaturedCarouselSlide = React.createClass({
```
blynch11p: Also no longer needed with the latest change

blynch11p:
Something to consider...

The current carousel does not apply the carousel features when the number of elements displayed can be shown on one screen (for example, bookmark as many Recommendations needed to remove all but 3 or 4 from the screen, and when the page is refreshed to remove them, the carousel arrows do not display)

blynch11p:
Works great!

The carousel is responsive.  Arrows only show when needed (and if the browser is resized, they show/hide as applicable).

Works in Featured, Recommended & New, and Listing Screenshots.

👍

app/js/components/carousel/index.jsx
```diff
@@ -2,76 +2,141 @@
...
-                    <div className="clearfix"></div>
-                </div>
-                <button className="prev btn hidden" href="#" ref="prev" aria-label="previous carousel button"><i className="icon-caret-left-white" alt=""></i></button>
-                <button className="next btn hidden" href="#" ref="next" aria-label="next carousel button"><i className="icon-caret-right-white" alt=""></i></button>
-            </div>
-        );
+    goToNextSlide: function (e) {
+        e.preventDefault();
+        var me = this;
+        var img_width = ((this.refs.slides.getDOMNode()).getBoundingClientRect()).width + 30;
```
irizvifearless: Instead of using a hard-coded value of 30, you can compute the space based on the margin of the slide component:

var slide = this.refs.slides.getDOMNode();
var gap = parseInt($(slide).css('margin-right')) + parseInt($(slide).css('margin-left'));

blynch11p:
👍

PR #915, opened 2017-12-11
fix(listing management): correctly sync data with table views on edit…

mleeBoeing:
Fix the table view not updating

Log onto Apps Mall as bigbrother

Go to Listing Management View | Table View ONLY

Choose Published Filter

Delete an app listing

RESULTS - Left side filter - Published  Filter total reduces by 1 and Deleted Filter increments by 1. (PASSED)  The Table View still displayed the deleted app while the Published Filter was selected(FAILED)

 

EXPECTED RESULTS - Left side filter - Published Filter total reduces by 1 and Deleted Filter increments by 1.  The Table View should NOT display the deleted app while the Published Filter is selected.

The actual results should still show the deleted app

### DIFF
--- a/app/js/stores/UnpaginatedListingsStore.js
+++ b/app/js/stores/UnpaginatedListingsStore.js
@@ -35,22 +35,25 @@ var UnpaginatedListingsStore = Reflux.createStore({
     onFetchAllListingsAtOnceCompleted: function (filter, response) {
         var key = filterKey(filter);
         var page = new PaginatedList (response);
-     
+
         _unpaginatedListByFilter[key] = page;
         this.trigger();
     },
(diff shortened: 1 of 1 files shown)


irizvifearless:
:+1: good fix

blynch11p:
Not working for me.

Log onto Apps Mall as bigbrother ✔️ 

Go to Listing Management View | Table View ONLY (My Listings) ✔️ 

Choose Published Filter ✔️ 

Delete an app listing ✔️ 

Published Count decreases ✔️  (deleted count not visible)

- Deleted Listing no longer has the Action icons
- Deleted Listing is still visible in table ❌

PR #918, opened 2017-12-19
refactor(PastNotification): refactor to use store

flesher1:
Refactors Notification to fetch the past store and notificaitons in a
method similar to active notifications.
Fixes skipped tests from loader modifications.
Solves store loading issues when making new listing notifications.
Fixes store loading issue when going from listing notification tab to
notification in mall management

AMLNG-857

**How to test**
Pull this branch. Check in listing modal > send notification
and in mall management > notifications


blynch11p:
Past Notifications for Listings and Center Settings are now correctly updated when switching between each.

👍

PR #919, opened 2017-12-21
feat(listing management): Visual Indicator Disabled

flesher1:
Makes disabled listings appear more opaque in this listing management
areas for better visual representation of what is enabled and disabled.

Closes #AMLNG-854

**Description**
As a user I would like a visual indication in Listing Management that a Listing is enabled or disabled.
Create and icon to display under a Listing in Listing Management if the listing is disabled.

**Acceptance Criteria**
- If a listing is disabled, an icon is visible under the Listing tile in the Listing Management Pages
- The icon does not replace the icon that indicated other status of the Listing, but is displayed near/next to that icon.

**How To Test**
Pull this branch. Go to Listing management. Disable a listing, it should be opaque. create a new listing and take it through the listing creation process. It Should be opaque until it is fully approved and enabled.

### DIFF
--- a/app/js/components/listing/ListingTile.jsx
+++ b/app/js/components/listing/ListingTile.jsx
@@ -187,17 +187,17 @@ var AdminOwnerListingTile = React.createClass({
         var imageLargeUrl = listing.imageLargeUrl;
         if(this.props.listing.approvalStatus !== 'DELETED'){
             return (
-                <li className={'AdminOwnerListingTile'}>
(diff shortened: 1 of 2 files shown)


mleeBoeing:
:+1:

blynch11p:
I would move the opacity: 0.4 to the <a> tag.  Otherwise all of the text, including the menu, become more difficult to read.

It also seems that after doing that in my testing, you'll have to apply a z-index to the <ActionMenu> to bring it back on top.

blynch11p:
better
👍

PR #923, opened 2018-01-03
feat(loading messages): display loading message for long running back-end tasks

irizvifearless:
closes AMLNG-863

**Description**
As a user, I would like to receive feedback when I send a notification, especially long-running system notifications.  When I send a notification, I would like to know that the request is being processed and what the outcome (success / failure) is.

**Acceptance Criteria**
- when the user sends a notification, the UI displays a message indicator informing the user of his/her action
- when the notification is successfully sent to all desired users or fails to be sent, the message indicator updates accordingly
- the user has the ability to hide the indicator in the UI

**How to Test**
Pull `notification_indicator` from `ozp-center`

It may be helpful to update the backend to simulate long notification requests and errors.  To do this, modify the file `ozp-backend/ozpcenter/api/notification/views.py`.  Add the following to the beginning of the `create` function of `class NotificationViewSet`:

```
    def create(self, request):
        import time

        # simulate 10 second notification
        time.sleep(10)

        # simulate getting an error
        raise Exception()

        # rest of the create function remains the same
        ...
```

### DIFF
--- a/app/js/actions/NotificationActions.js
+++ b/app/js/actions/NotificationActions.js
@@ -5,15 +5,57 @@ var createActions = require('../utils/createActions');
 var NotificationApi = require('../webapi/Notification.js');
 var ProfileActions = require('ozp-react-commons/actions/ProfileActions');
 
+var MessageStore = require('../stores/MessageStore');
+
+var _notificationTypes = [
+    {key: 'System', value: 'System'},
+    {key: 'StewardAppNotification', value: 'Update Request'}
+];
+
 var NotificationActions = createActions({
 
     createNotification(uuid, notification) {
+        var displayType = ' ';
+        if(_.has(notification, 'listing')) {
+            displayType = ' Listing ';
+        }
+        else {
+            var notificationType = _.find(_notificationTypes, { key: notification.notificationType });
+            if(notificationType) {
+                displayType = ' ' + notificationType['value'] + ' ';
+            }
+        }
+        var messageData = MessageStore.addMessage('Sending' + displayType + 'Notification', true);
+
         NotificationApi.create(notification)
             .then(function (response) {
                 NotificationActions.createNotificationCompleted(uuid, response);
+                MessageStore.updateMessage(messageData['uuid'], {
+                    message: messageData['message'] + ' Complete!',
+                    state: MessageStore.SUCCESS_STATE,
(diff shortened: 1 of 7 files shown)


irizvifearless:
For now, I use two different approaches to give the user feedback: 

1) A message that displays near the Send button in Center Settings > Notifications.  The message stays there until the request finishes, and in the event of an error, shows the error until the next request to send a notification.  The reason for that is to show the user the status if their request, even if they navigate away from the Notifications tab, then come back.

2) An alert that displays on the bottom of the screen.  The alert updates when the request finishes (success/failure).  The user has the option to hide the alert.

Right now, the two approaches are used simultaneously.  Any feedback/comments on the approaches is appreciated!

Currently, the limitations to this approach are as follows:

1) If the user forces a page reload while the notification request is being processed, all ajax event handlers, reflux action listeners, etc are destroyed.  The `.fail()` functions are immediately called when the window unloads, simulating an error (luckily, I can differentiate this from an internal server error), and the success handlers are never called, even if the response returns with status code '200'.  Unfortunately, clicking the center button on the nav bar forces a page reload.

For now, I just hide the alert / message when that happens.  In order to give the user proper feedback after page reload, one option would be to long poll the backend until the notification request completes (cumbersome, inefficient solution but gets the job done).  Another would be to persist the alert / message to the DB (right now, it's persisted to session storage), but that would require creating a DB model, view, serializer, api, and unit tests on the backend, then creating an api hook, actions, updating the store, etc on the frontend (major effort).

Any other thoughts/recommendations on this issue?

Finally, I tried to make the MessageQueue component as flexible and extensible as I could, so it _theoretically_ can be used for other API calls or actions.  Please test and let me know if you see ways to improve it.

blynch11p:
Initial review:
I like the banner at the bottom notifying the user of success. It looks good, with good details.
We probably don't need to require user interaction.  We can leave the X but have it close after ~ seconds.
The banner is behind the overlay when a Listing is open, and also behind the listing when the browser is minimized.

irizvifearless:
Updated so that whenever you cause an action to add a new alert message, all finished alert messages are cleared.  Let me know if you like this approach or if it would be better with the ~5 timeout clear.

flesher1:
Works great. Looks good. :+1:

PR #926, opened 2018-01-18
fix(carousel): Styling fix

flesher1:
Add styling to support older browsers


blynch11p:
Tested in Production (actually, Staging)
👍

mannyrivera2010:
:+1:

PR #927, opened 2018-01-18
fix(listing management): make toggling between list and table view consistent through all tabs

irizvifearless:
closes AMLNG-892

**Description**
In listing management, when a user toggles between list view and grid view, their state is not always preserved across tabs

**Acceptance Criteria**
Log in as an Org Steward (ie julia) -- you can test with a variety of users, but make sure to test with at least one org steward that has an org listing tab
Go to Listing Management
Toggle between list and table view
RESULTS: Changing tabs does not preserve the list/table view selection
EXPECTED RESULTS: Changing tabs preserves the list/table view selection

**How to Test Code**
Pull `inconsistent_state_fix` from `ozp-center`

### DIFF
--- a/app/js/components/management/AllListings/index.jsx
+++ b/app/js/components/management/AllListings/index.jsx
@@ -27,7 +27,7 @@ var AllListings = React.createClass({
     ],
 
     getInitialState: function () {
-        var useTableView = JSON.parse(sessionStorage.getItem('center-allListings-toggleView'));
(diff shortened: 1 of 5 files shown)


flesher1:
:+1:  Tested and reviewed.

`Uncaught Error: Invariant Violation: replaceState(...)` is thrown when switching from org to my-listings tabs. Checked with Irtiza, error determined to be on master as well. Will be addressed separately.

PR #928, opened 2018-01-23
fix(listing management): deleting a listing correctly updates the table view

irizvifearless:
closes AMLNG-891

**Description**
In listing management, when a user deletes a listing while in table view, the listing still shows up in the list, even if the status filter is not set to Deleted.  The listing is displayed as deleted, but should not be a part of the list.

**Acceptance Criteria**
Go to Listing Management
Toggle to show the table view
Set the sidebar filter to "Published"
Delete a listing from the table view
RESULTS: Listing displayed as deleted, but shows up in the list
EXPECTED RESULTS: Listing is no longer a part of the "Published" filter and is in the "Deleted" filter

**How to Test Code**
Pull `pending_delete_fix` from `ozp-center`

### DIFF
--- a/app/js/components/carousel/index.jsx
+++ b/app/js/components/carousel/index.jsx
@@ -126,7 +126,7 @@ var Carousel = React.createClass({
         );
 
         return (
-            <div className="carousel" onResize={() => console.log('resized!')}>
+            <div className="carousel">
                 {showArrows ?
                     <div>
                         {items.length === 1 ?
--- a/app/js/components/management/shared/TableView.jsx
+++ b/app/js/components/management/shared/TableView.jsx
@@ -383,15 +383,14 @@ var TableView = React.createClass({
         }
 
         this.props.onCountsChanged(counts);
-
-        // TODO: Figure out better way for counts updating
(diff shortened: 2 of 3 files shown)


irizvifearless:
Although this approach correctly syncs the table view with the updated listing data (also removes hack to update counts after a listing is deleted), it clears the table's cached results whenever the listing is updated, which may cause a performance degradation.

Let me know if this is a problem.  The alternate approach is to manually update the data cached in the front end, but this is susceptible to data being de-sync'd between the front and back end.

blynch11p:
This makes sense, and from a user standpoint, I don't think it's an issue to have a complete refresh when you're messing with the table.

I think the 'deleting from All Center Listings, and ending up on the My Listings page' is a separate issue from this, yes?

:+1:

irizvifearless:
Yes.  Did some research and it looks like that stems from the `close` function of the `ListingDeleteConfirmation` component in `shared/DeleteConfirmation.jsx`.  Manual redirect to the my listings view whenever the listing delete is successful.

flesher1:
Works well, code looks good.
:+1: 
Thanks for removing my console.log I forgot about.

PR #937, opened 2018-02-08
Older browsers don't support just column-count, need to add the -moz …

blynch11p:
…and -webkit variations

Another hard one to test...

`column-count` works on the current Chrome/FF, but not older browsers.

Adding -moz & -webkit versions appear to fix the column layout for Usage / System requirements, so they are displayed in two columns instead of one.


flesher1:
:+1: looks fine from here.

irtiza-rizvi:
:+1: works on browsers I'm able to test with

PR #939, opened 2018-02-09
feat(listing): require feedback when deleting a listing

irtiza-rizvi:
AMLOS-545

**Description**
I would like to capture the reason when an app steward deletes a listing by requiring the app steward to submit a reason during the deletion process.

**Acceptance Criteria**
1. include feedback form in the deletion process
2. require the app steward to provide feedback in order to complete the process

**How to test code**
Checkout `delete_listing_feedback` from `ozp-center`
Checkout `delete_listing_feedback` from `ozp-backend`

_Test Scenario 1_
Log in as bigbrother
Delete a listing via Listing Management
Ensure that you are required to enter a reason for deletion
Ensure that the reason displays in Listing Management > Recent Activity

_Test Scenario 2_
Log in as jones
Pend a listing for deletion
Log in as bigbrother
Go to Administration tab of listing
Approve listing for deletion
Ensure that you are required to enter a reason for deletion
Ensure that the reason displays in Listing Management > Recent Activity

### DIFF
--- a/app/js/components/quickview/AdministrationTab.jsx
+++ b/app/js/components/quickview/AdministrationTab.jsx
@@ -111,7 +111,16 @@ var AdministrationTab = React.createClass({
 
     getInitialState: function () {
         PaginatedChangeLogByIDStore.resetChangeLogByIDStore();
-        return { prevId: this.props.listing.id, editingReviewRejection: false, editingDeleteRejection: false, hasMore: true, changeLogs: [], loading: false, loadingError: false  };
+        return {
+            prevId: this.props.listing.id,
+            editingReviewRejection: false,
+            editingDeleteRejection: false,
+            editingDeleteApproval: false,
+            hasMore: true,
+            changeLogs: [],
+            loading: false,
+            loadingError: false
+        };
     },
 
     componentWillMount: function(){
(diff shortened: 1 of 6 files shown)


flesher1:
:+1: 
Checked the direct admin delete. Works as expected.
Checked the pending delete process. Works as expected.
Checked org steward. It worked for them too.

app/js/components/quickview/AdministrationTab.jsx
```diff
@@ -322,21 +332,33 @@ var AdministrationTab = React.createClass({
                     </form>
                 </section>
             );
+        } else if (editingDeleteApproval) {
+            return (
+                <section className="return-feedback">
+                    <h5>Approve Deletion Feedback</h5>
+                    <p>Please provide feedback for the listing owner about why their listing should be deleted.</p>
```
blynch11p: The text here could probably be reworded.  The owner already pended the listing to be deleted, so they don't need to be told why it should be.  Maybe just use what's in the pop up dialog:
"Please enter a reason for deleting the listing "

irtiza-rizvi: Good idea.  I've pushed an update with that change.

blynch11p:
Test scenarios pass. Text was updated.

👍

PR #940, opened 2018-02-13
fix(listing management): use org short name when filtering by org

irtiza-rizvi:
Fix to a bug created by https://github.com/aml-development/ozp-center/pull/925

When I refactored the organization tabs in Listing Management to use the short name instead of the full title in the URL, it broke the organization filter in the sidebar.  This update fixes that issue and closes AMLNG-801.

Special thanks to @mannyrivera2010 for uncovering this bug!


mannyrivera2010:
:+1:

flesher1:
:+1:

PR #942, opened 2018-03-14
fix: fix review time posted

mleeBoeing:
fixes review showing a different timestamp when editted. 

How to test:
1. Login and post a review on a certain listing.
2. Logout and login as a different user and post a review on the same listing above.
3. Logout and log back in with the account you used in step 1.
4. Edit the original review you have created. Time shouldn't change nor position.


flesher1:
:+1:

PR #948, opened 2018-11-27
fix(accessability): add dynamic change notificaiton

flesher1:
Adds notification to user when changing to search page, or returning to 
home page.

Closes #AMLOS-927


mleeBoeing:
:+1:
Looks good but currently do not have a JAWS machine to test

